PR #2, opened 2016-02-17
Files with "_" in their path could not be handled right

xolotl90:
Files with "_" in their path could not be handled right


tunnell:
Great, thanks.

PR #11, opened 2016-04-08
Added a detector option in runs.py to be able to use hax with muon_veto

CyrilGrignon:


hax/runs.py
```diff
@@ -51,8 +51,13 @@ def update_datasets():
...
         db = client[hax.config['runs_database']]
         collection = db[hax.config['runs_collection']]
+        if 'detector' in hax.config:
+            detector = hax.config['detector']
+        else:
+            detector = 'tpc'
+
         docs = []
-        for doc in collection.find({'detector' : 'tpc'},
+        for doc in collection.find({'detector' : detector},
```
tunnell: __[flake8]__
* [E203] whitespace before ':'

tunnell:
The changes look good.  After that small tweak, commit.

PR #13, opened 2016-04-12
Peak properties

tunnell:
I propose we add a TreeMaker for peak properties.  I'm not sure if we want one tree maker for all types or can say 'S1PeakProperties' etc.

Thoughts?


hax/treemakers/common.py
```diff
@@ -101,3 +101,54 @@ def extract_data(self, event):
...
+            peak = event.peaks[index]
+            # The store each peak field we want in 'values'
+            for field in self.extra_branches:
+                field = field[6:]
+                field_name = '%s_%s' % (name,
+                                        field)
+                values[field_name] = getattr(peak,
+                                             field)
+
+        return values
```
tunnell: __[flake8]__
* [W292] no newline at end of file

PR #29, opened 2016-06-14
Hack dataextractor to get xy for a peak

JelleAalbers:
This is a small hack to the DataExtractor so we can get the xy for each peak. Erik, do you think there is a better way, or is this fine for now?


ErikHogenbirk:
Looks good and very useful indeed, if there are other fields like range_50p_area and x,y (i.e. fields that cannot be added in the standard way of peak.property) we can add them in a similar way.

PR #30, opened 2016-06-14
Pax version checks towards minitree standardization

JelleAalbers:
This improves how hax deals with data processed by different pax versions at analysis facilities, to facilitate minitree standardization.

 * Hax will use the runs db to locate the pax root file with the right version (for what the 'right version' is, see below).
 * Minitrees now have a pax_version key in their metadata, specifying which pax version created the original big root file. Hax will not load minitrees if that version is wrong, but instead create a new one (from the root file with the right version).

The option pax_version_policy controls what versions are 'right' and 'wrong'. Your options are:

 - **latest**: always get data from the latest pax version available for a dataset. Pro: this works for every dataset for which some data is available. Con: it is dangerous for analyses on many datasets (as you risk mixing data from different pax versions) and analyses whose results should be reproducible later (as your results may change without notice once pax gets updated).
 - **X.Y.Z** (with X, Y, Z numbers): only get data processed with pax version X.Y.Z. Pro: You are not susceptible to the risks above. Con: You can't analyze datasets which have not (yet/anymore) been processed with your chosen version.
 - **loose**: The current behaviour. The runs database is not used to find data locations, instead we look for whatever root files exist in main_data_paths. Minitree version checking is also disabled. Pro: works outside analysis facilities, without the runs db, for XENON100, secret TPCs in attics, root files people created themselves (e.g. simulated data), and other strange conditions. Con: you are completely responsible for ensuring the data used by hax is processed by the pax version you want.

Which policy you want depends on the kind of analysis, the availability of data from different versions, how fast the pax versions change, your personal preferences, and eventually probably what your analysis coordinator tells you to use :-)

The default policy is 'latest', for two reasons:
  * The default should work out-of-the box on all datasets, including datasets which people make on their own;
  * A default of X.Y.Z would mean we have to update that default regularly. If people rely on this (changing) default, their analysis results would still change without notice!!

If you want to compare different pax versions on the same data, call hax.init() a second time after you load the first version, specifying the second version in the version policy.

Even though minitree versions are checked, currently the minitree finder still just selects the first minitree it finds. If that's not the right version, a new one will be created. Once we've standardized a minitrees location, and you want to use non-standard-version-minitrees, eg from an old pax version, you can explicitly set e.g. minitree_paths=['.', '/path/to/old/minitrees']. 

If you are not at an analysis facility, you have to explicitly set pax_version_policy='loose', to acknowledge you are aware you're not getting any version consistency checks. Otherwise you get this error:

    >>> hax.init()
    DEBUG:hax.__init__] Read in hax configuration file /home/jelle/hax/hax/hax.ini
    DEBUG:hax.runs] Updating datasets from runs database... 
    DEBUG:hax.runs] ... done.
    DEBUG:hax.runs] Folder /mnt/lena/xenon1t/ not found when looking for raw data
    WARNING:hax.__init__] You're not at a XENON analysis facility, or hax can't detect at which analysis facility you are.
    Traceback (most recent call last):
      File "<stdin>", line 1, in <module>
      File "/home/jelle/hax/hax/__init__.py", line 66, in init
        raise ValueError("Outside an analysis facility you must explicitly set pax_version_policy = 'loose', "
    ValueError: Outside an analysis facility you must explicitly set pax_version_policy = 'loose', to acknowledge you are not getting any version consistency checks.

I tested this on midway (using my own anaconda), seems to work.


JelleAalbers:
We might consider putting the default to 'loose' for the moment, so we don't break people's existing code. If so, we should probably give a warning we eventually want to change the default.

JelleAalbers:
I'd like to merge this soon to avoid more minitrees without version info being created. Any objections or opinions?

JelleAalbers:
If there are no objections I intend to merge this tomorrow to avoid more and more minitrees without version info being created.

PR #31, opened 2016-07-20
Improvements for basic treemakers

JelleAalbers:
~~This splits the Basics treemaker into two: Basics and MainInteraction. Basics now contains a 'bare minimum' of information: the event time, duration, and run number. It loads really really fast. MainInteraction contains everything else, and now includes also the s1 and s2 width (range_area_decile[5]).~~

~~Both Basics and MainInteraction are loaded by default if you just call hax.minitrees.load, but if you did this: hax.minitrees.load(run_number, ['Basics', MyTreemaker]) and use the main interaction properties, you have to change it to: hax.minitrees.load(run_number, ['Basics', 'MainInteraction', MyTreemaker])~~

Please see the posts below for an updated description.

The main reason for this change is that not all analyses need the full main interaction info (e.g. LargestPeakProperties analyses, analysis on trigger signals, ...), but you always need the event number and run number to inspect events. 

This also fixes the LargestPeakProperties treemaker and a small bug with inspect_events_from_minitree.

### DIFF
--- a/hax/__init__.py
+++ b/hax/__init__.py
@@ -10,7 +10,8 @@ __version__ = '0.4.1'
 # I am surprised this works (even if we do 'from hax' instead of 'from .')
 # as some of the modules do 'import hax' or 'from hax.foo import bar'.. shouldn't we get circular imports??
 # I need to read up on python packaging more...
-from . import ipython, minitrees, paxroot, pmt_plot, raw_data, runs, utils, treemakers, data_extractor
+from . import ipython, minitrees, paxroot, pmt_plot, raw_data, runs, utils, treemakers, data_extractor, \
+              slow_control, trigger_data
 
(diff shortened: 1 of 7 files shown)


JelleAalbers:
Actually, a better way may be to rename the new "Basics" to "Fundamentals" and load it always no matter what, then we can keep the old "Basics" and people's code will continue to work.

JelleAalbers:
I've changed things per the suggestion above: there is now a 'Fundamentals' treemaker that is silently loaded whether you like it or not. The Basics treemaker now still provides the info about the main interaction info.

This now also adds:
 *  `hax.trigger_data.get_trigger_data`, a function to get the trigger monitor data from the zip file. 
 * A new treemaker called LargestTriggeringSignal, which provides information on the largest trigger signal with the trigger flag set to True.

PR #32, opened 2016-08-25
Minitree save arrays and pickle

jhowl01:
# Changes in this branch
There are a couple changes to minitrees.py in this branch, and one new jupyter notebook containing a usage example ("minitree_with_arrays"). Here I will explain the motivation for these changes.

## Pickling vs Root
The main hax output is currently a pandas DataFrame which is returned by `minitrees.load()`, as well as a `.root` file produced from this DataFrame and saved in the `minitree_paths` directory. While one can use the returned DataFrame for pandas analysis, some prefer to do further analysis using the reduced saved dataset.

Since pickling allows one to serialize a pandas DataFrame without bringing root into the mix at all, we thought it may be useful to be able to save in this format as well or instead, especially since some people prefer to do analysis without root.

**Pickle -** Analysis with pandas/numpy/matplotlib

**Root -** Analysis with root, backward compatibility

## Saving arrays in a minitree
Currently, minitrees are flat, in that each branch takes a single value per event. If you make a TreeMaker with an `extract_data()` function that returns a dictionary containing arrays, the array-type fields will not show up as branches in the output `.root` file, nor will they appear in the returned DataFrame. This is fine for many applications, but becomes difficult if you want to look at more than a few interactions per event (e.g. peak- and hit-level analyses).

### DataExtractor vs minitrees with arrays
Hax now includes the `DataExtractor()` class for analyses of this type. Its output is a numpy structured array which is also flat in that none of its columns contain arrays, but it does allow one to save information about _all_ of the peaks or hits in a dataset.

**DataExtractor Output**
dataextractor_table

**Minitrees w/ Arrays Output**
branch_table

There are a few reasons some might prefer to use a minitree with arrays rather than DataExtractor to look at peak-level info. First, the array branches are space-saving, since in the DataExtractor output, the values of each event-level field are repeated for each peak in an event. Also, the DataExtractor method does not save the results, and if they were saved to a `.root` file it would not have the usual structure where each entry is an event and peak information is represented as array-type branches. For these reasons, the minitree method may be ideal for those of us who use hax to produce a reduced copy of the processed `.root` file on midway, then do our analysis on the reduced file wherever we want.

It may also be advantageous that this branch contains no change on the user side, except that array-type fields no longer disappear - so if an analysis using minitrees expands to the point where the user prefers arrays to many dictionary keys, this can be accommodated without any change in code. Note that the new function `dataframe_to_root()` is only used if the dictionaries returned by `extract_data()` contain arrays, and otherwise there is no change except those made to allow pickling.

## Structure of changes
minitree_edits 001
Basically, rather than automatically passing the DataFrame to a `.root` using `root_numpy.array2root()`, and then retrieving it using `root_numpy.root2array()` before returning it, in my branch the DataFrame is returned directly. First though, it checks for the options `save_root` (defaults true) and `save_pickle` (defaults false). If `save_root` is true, it will check if the dictionaries contain arrays. If they do, it will use my new function `dataframe_to_root()`, and if not it will use `array2root()` as normal. If `save_pickle` is true, a dictionary containing both the DataFrame and the same metadata dictionary will be saved to a `.pkl` with the same header.


tunnell:
Hey, thanks for this!  It looks interesting.  Our thought was always that people would do Hit/Peak analysis on the pax rootfile or create one row per peak in the dataframe (i.e. no interaction info).  What you propose in interesting.  Can you check one thing? One issue we always had was that Pandas doesn't easily let you do operations on arrays of values and as far as I know the only ways to do this are ROOT/MongoDB query languages.  Are you able to plot all peaks for one interaction this way?

You might need to do a list comprehension or something per row in the table over the array to create the value that you're interested in analyzing.

JelleAalbers:
Hi Joey,

It's great to see someone new has read the code, actually understood it, identified a problem, then contributed something useful. You also explain it in a clear way so it's easy to review. The code looks quite readable, and the example notebook also helps understand what's going on. So thanks a lot!

The array fields feature address a big problem several analysts are having. DataExtractor exists, but it's interface is less nice, it lacks some features (most importantly file-based caching) and has less worked examples, so people don't often use it yet. Instead, I've seen people packing their stuff in multiple columns (interaction_0, interaction_1, ...) of a minitree... :-(

Still, I think DataExtractor is what people usually want, even though they don't know it yet :-). Take a single electron analysis. With your feature I can collect properties of some peaks (say area, width, x, y) in array fields in a minitree. But after I made the minitree, how would I select, say, only peaks with an area larger than 10?  Remaking the minitree for every selection is very slow. I think I would have to iterate over the rows, get a boolean mask from the area field, then apply it to all of the peak property fields in the array. 

If this makes people's life easier we should definitely merge it, but I'm worried it's actually quite difficult to work with once you go beyond just creating a minitree and looking at it. Maybe you can give an example of how you would use this in your analysis?

About save_pickle: any move away from formats used only in the particle physics bubbleverse of course has to be welcomed :-) In pax we use pickle as the raw data format, but we compress the pickles with zlib first. I'm not sure if compression is needed for minitrees, but maybe if we start storing arrays in them they will become large. Might be worth a look.

Am I correct though, that hax itself still only reads from the .root cached minitrees? So you can save to pickle, but then have to deal with reading it in, concatenating it with other minitrees etc. yourself. Then I'm not sure why you would like to have this in hax.minitrees. Doesn't really hurt anyone if it's just an option, but...

Thanks for making this pull request, again I really appreciate that you just dove into this and came out with something useful.

hax/minitrees.py
```diff
@@ -161,16 +218,25 @@ def get(run_name, treemaker, force_reload=False):
...
+        empty_frame = pd.DataFrame()
+        return minitree_path, empty_frame
 
     # We have to make the minitree file
     # This will raise FileNotFoundError if the root file is not found
     skimmed_data = treemaker().get_data(run_name)
+    
+    # Setting save_arrays to True if any arrays/vectors in DataFrame
+    for branch_name in list(skimmed_data):
+        if hasattr(skimmed_data[branch_name][0], "__len__"):
```
JelleAalbers: Strings and bytes also have the __len__ attribute, but you probably don't want them to trigger your array saving logic. You could broaden the check to
```
x = skimmed_data[branch_name][0]
if hasattr(x, "__len__") and not isinstance(x, (str, bytes)):
    ...
```
This will still raise a rather scary/uninformative IndexError if there is no data in skimmed_data -- but that's a very rare edge case. There's a another instance of this check in dataframe_to_root (might be worth factoring out to a function).

hax/minitrees.py
```diff
@@ -161,16 +218,25 @@ def get(run_name, treemaker, force_reload=False):
...
+            save_arrays = True
```
JelleAalbers: Let's print / log a warning rather than just  turning this on silently. Maybe say something like "Branch %s is an array field: turn on save_arrays when making minitrees, or use DataExtractor.". Personally I would even throw an error rather than a warning, so the user will turn it on themselves, but that's up to you.

jhowl01:
Hi guys, thanks for the quick responses!

First to the general comments, I've come to agree with both of you - at first I saw minitrees didn't take arrays and was used to xerawdp, but I've since been working on this in parallel with doing single electron stuff with the DataExtractor, and agree that it is better in this case. I think the idea here is that its a more familiar structure to people, and could be preferable in some cases, like if an existing analysis expands to be looking at many peaks per event, but still on an event-by-event basis. I also saw examples with separate s2[0] and s2[1] columns in a minitree, and thought it would be nice to be able to easily look at the first 'n' s2s without a different class.

@tunnell It seems you're totally right that although root can do this, in pandas its not possible to go directly to, say, a plot of two columns for all peaks in the dataset without some sort of loop over the rows, which I realize is time-consuming. I think for an analysis like this the DataExtractor is better.

@JelleAalbers I actually had only been looking at one dataset at a time (with force_reload turned on), so I hadn't even thought about loading and concatenating existing trees with hax. This was an oversight I will look at - it shouldn't be hard to also be able to load and concatenate existing pickled minitrees, but I think to go from a "minitree with arrays" root file to a dataframe with array-type columns, I will also need a new loading function (e.g. root_to_dataframe). I agree with your other comments and will implement them.

I'm realizing now its a bit confusing that there's really two changes being requested here, pickling of minitrees and minitrees with arrays. I will implement these changes and commit them to the branch, and then we can decide if one or both of these modifications is worth merging in some form.

Thanks a lot for the encouragement guys, getting into the hax source code like this has been a great crash course in xenon1t analysis.

tunnell:
Did you consider multiindex pandas dataframe instead of arrays?

jhowl01:
I hadn't thought of that cause this all started from a root-y place, but its a good idea and more to the point of what I'm trying to do. I can try having the returned dataframe/pickle have this format, and the saved rootfile keep the array-type branch form.

jhowl01:
Hey Guys,

Most of the changes discussed have now been committed to the branch. First, the options are now `'use_root'` (default True), `'use_pickle'` (default False), and `'use_arrays'` (default False). Second, all minitree output options can be loaded and concatenated in the usual way if `'force_reload'` is left off.

Here's a little truth table of the options (assuming `force_reload=False`)

|  `'use_root'`  | `'use_pickle'`  |  Result  |
| :--------------: | :----------------: | :----------------: |
|  True  |  False  |  (DEFAULT) Saves as root, loads from root |
|  True  |  True  |  Saves as root and pickle, loads from root |
|  False |  True  |  Saves as pickle, loads from pickle  |
| False | False  |  Saves nothing, loads from root  |

@tunnell,
I'm having some issues with using MultiIndex DataFrames instead. Unless I'm mistaken, it seems that every column is required to have the same dimensionality, leaving something like the DataExtractor output but with smarter indexing allowed (the first two columns are indices):

| | |  peak_area  |  event_time  |
|---:|---:|---:|---:|
| event | peak | | |
| 0 | 0 | 3.8 | 1472671002 |
| 0 | 1 | 15.1 | 1472671002 |
| 0 | 2 | 7.5 | 1472671002 |
| 1 | 0 | 4.5 | 1472671009 |
| 1 | 1 | 10.0 | 1472671009 |
| 2 | 0 | 20.5 | 1472671083 |

However, it does allow one to manipulate all peaks without a loop, as you pointed out my method couldn't - for instance:
`in [0]: np.asarray(df['peak_area'])`
`out[0]: [3.8, 15.1, 7.5, 4.5, 10.0, 20.5]`

Maybe this is worth considering for DataExtractor?

I will update the example notebook included as well.

\- Joey

JelleAalbers:
Hi Joey,

Thanks for implementing all the suggestions; I think it's ready to merge now. Looks like some things have changed in master so there's a conflict, but I'll resolve it during the merge.

Erik and I are working on essentially integrating dataextractor in minitrees, so you can make peak-per-row treemakers. This doesn't make your work obsolete though -- if you want to save a list field (such as area_per_channel) you'll need to use either use your save_pickle or the dataframe_to_root function. 

Perhaps we can make use_arrays / use_pickle a hax.init option instead, and make use_arrays a treemaker attribute (since a treemaker either always or never needs use_arrays). 

Thanks again for your work!

Jelle

JelleAalbers:
While preparing the merge I came across a few things that might have caused trouble and changed them. @jhowl01 maybe you could have a look at it to see if you agree before we merge?

 * Instead of a `use_arrays` option to `minitrees.load`, I added an attribute `uses_arrays` to the TreeMaker base class. When you set this to True (default False) when defining yourr treemaker, it will triggers your array saving code *only for that treemaker*. This matters, as your array saving code might produce slightly different root files than the standard root_numpy code. If we have `use_arrays` as an option to `minitrees.load`, somebody will write `minitrees.load(run_id, ['Basics', 'SomeArrayFieldUsingTreemaker'], use_arrays=True)` and end up with a different Basics root file than somebody that doesn't use `use_arrays`. 

 * I also changed the 'length branch' stuff:
  * You asked the user to supply, for each array column, an additional column (with an arbitrary name) containing the lengths of each array field. To match array columns to "length columns", you checked, for every array column, every other column in the dataframe to see if its first ten elements are the lengths of the arrays in the array column. This can fail in many ways -- e.g. if dataframe does not have ten elements or if an array field is usually empty and there are several other columns that are usually/always 0. 
  * Instead I required the length column to be called `x_length` where x is the name of the array field. The column is also automatically added to the dataframe (only when your save arrays logic is triggered).
  * I was wondering if we could use vector branches to store the array fields instead, then you wouldn't need the lengths, right? But I don't suppose it matters much.

Here is a minimal example. Run `paxer --config XENON1T Simulation --output test_output` to create a small root file, then

```
import numpy as np
import hax
hax.init(pax_version_policy='loose', main_data_paths=['.'])


class TestList(hax.minitrees.TreeMaker):
    uses_arrays = True
    __version__='0.0.1'
    
    def extract_data(self, event):
        return dict(a=np.random.randint(0, 10, size=np.random.poisson(10)))
                       
hax.minitrees.load('test_output', TestList)
```

jhowl01:
Hey Jelle,

Sorry for the delay, classes have just started up again here. Thanks for making these changes, they make good sense. The merging of the two classes is a great idea! This should help some of the change-averse to adopt DataExtractor.

- As for the 'length branch' issue, I realized there were issues with the check I was doing but wasn't sure how to handle them - this is much better but the only issue is that several branches could be made which are identical except for their names ('peak_area_length', 'peak_width_length', etc.). I'm not sure how best to handle this.

- I had a version of this working with stl vectors in rootpy instead, and as you guessed this didn't require length branches, but I found it to be much more time-costly. The fastest I could get it, the saving step took around the same amount of time as the loop_over_dataset step. This was why I went the arrays route.

- Ultimately I agree that making these hax.init options would be better, unless someone wanted to make both types of minitrees in a single load line, which I would imagine to be a very rare case.

- Joey

JelleAalbers:
Hi Joey, thanks for your feedback, I've merged the pull request with master now. 

Thanks for trying out the vector branches, I I tried something like this for some of the array fields in the pax root file output too (e.g. sum waveform and area_per_channel), and seem to remember that was also slow. 

I think with the new peak-per-row minitrees this feature will be most often used to get one or a few array fields (such as area_per_channel or sum_waveform), so duplication of length branch information is hopefully not a big issue. But I had not considered this, that helps explain why you originally went for the 'compare first ten items' solution.

PR #33, opened 2016-09-08
Multiple row minitrees

ErikHogenbirk:
This commit introduces minitrees with multiple rows per event, which is useful for reading in data of all sub-event properties, such as peaks, hits, interactions, etc. In this commit, only peak information is implemented completely (but other extensions are straightforward). I added a full example with an explanation of all options and some caveats. 

This pull request will enable us to incorporate the previously separate DataExtractor into the TreeMaker framework. Will greatly simplify any peak-based analysis, imo.


JelleAalbers:
Hi Erik, Thanks a lot for this! I hope this will make peak-level analyses more accessible to people. The example notebook looks good; it seems the single electron area has gone down somewhat... can't really tell from the scatter plot. But that wasn't the point of the example of course.

One small comment: can you remove the `__version__` attribute from the PeakExtractor base class and add a `__version__` attribute to the example custom classes you show in the notebook? Right now if I make a class based on PeakExtractor, it will always have version 0.0.2 and I can't update it. One of the benefits of minitrees is that you can bump treemaker version numbers, so outdated minitree files will be remade.

A point for people to keep in mind is that attributes you specify to custom `PeakExtractor`'s (such as `stop_after` or `peaks_cut_list`) affect the behaviour of the treemaker. If you change them, you have to bump the treemaker version number (or for development, use `force_reload=True`) to ensure your minitrees get remade. Otherwise your changes will seem to do nothing after a minitree has been made and cached.

I've worked with you on this, so it already passes the four-eyes test; I don't think it needs a detailed review beyond that.

PR #36, opened 2016-10-05
DoubleScatter treemaker updated

berget2:
The search for double scatter events:
        double decays, afterpulses, and anything else that gets in our way
        if you have any questions contact Ted Berger (berget2@rpi.edu)
        
    The search proceeds as follows:
      * interaction[0] (int_0) provides s1_0 and s2_0
      * find additional interaction (int_1) to provide s1_1 and s2_1
        - loop through interactions (interactions store s1/s2 pairs in 
          descending size order, s2s on fast loop)
        Choice A) select first interaction with s1 != s1_0 AND s2 != s2_0
        Choice B) if Choice A doesn't exist, select first interaction with 
                  s1 != s1_0 AND s2 == s2_0
        Choice C) if Choice A and B don't exist ignore, this isn't a double 
                  scatter event
      * int_0 and int_1 ordered by s1.center_time to int_a and int_b 
        (int_a has s1 that happened first)
   
For a full description of the output variables please see the comments in the t

### DIFF
--- a/hax/treemakers/DoubleScatter.py
+++ b/hax/treemakers/DoubleScatter.py
@@ -1,0 +1,268 @@
+import hax
+import numpy as np
+
+class DoubleScatter(hax.minitrees.TreeMaker):
+    
+###############################################################################
+    
+    """
+    The search for double scatter events:
+        double decays, afterpulses, and anything else that gets in our way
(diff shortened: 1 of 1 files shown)


JelleAalbers:
Thanks Ted! The documentation will help figure out what it does, and this is already used by several people, so let's merge it. We can always update it later.

PR #37, opened 2016-10-07
Parallel minitrees, preselections, and out-of-core computations

JelleAalbers:
This includes several enhancements to prepare hax for large-scale analysis. 

Here is an [example notebook](https://github.com/XENON1T/hax/blob/outofcore/examples/Preselections%20and%20out-of-core%20computation.ipynb).

Parallel minitree loading
-------------------------
`minitrees.load` now has a `num_workers` option, which will spread the loading of several minitrees over multiple cores Each core gets separate runs, runs are not split over cores.

This is useful to create minitrees faster from the root files, and works well with the preselection feature (see below).

Please don't abuse this option to spam the jupyterhub with hundreds of minitree creation processes when people want to use it for analysis. If you need many minitrees, use a midway compute node / batch queue job, ask your friends to share their minitree folders -- or better yet, lobby the computing group to maintain a folder with pre-made standard minitrees!

The multiprocessing is done by the dask multiprocessing scheduler. If you have a remote machine with a [dask scheduler](http://dask.pydata.org/en/latest/scheduler-overview.html) setup you can try to use that by passing `dask_compute_kwargs=dict(get=your_scheduler)`. I think. I haven't tried this.

Preselections
-------------
Often you're interested in a small part of the data, e.g. the low-energy events. You can then use the preselection option:

    data = hax.minitrees.load(run_numbers, preselection='cs1 < 1000', num_workers=10)

This combines particularly well with parallel minitree loading, since the preselection can then be applied separately on each run, i.e. "out of core". Previously you had to first load in the full data -- potentially blowing up your RAM -- before you could apply a cut.

Note that 'cs1' is just a variable defined by the basic treemaker; you can use any other variable defined by a treemarker you're loading in. So if you have more complicated preselection cuts, use a treemaker to output a cut boolean (or p-value, or whatever you want to cut on) and use that in the preselection string. The preselection is evaluated for every run after merging the dataframes for different treemakers.

The cut history of your preselection is kept, as if you did a normal cut. (this required some strange magic for the parallel case)

Some finer details:
  * You can specify a string or a list of strings that can be parsed by pandas.eval. 
  * Passing a list of strings is the recommended way to pass multiple preselections (instead of using & to string conditions together). This way we can track the cut history separately.
  * If you need to refer to e.g. the local x variable, also pass `pandas_eval_kwargs=dict(local_dict=dict(x=x))`, then you can say `preselection = cs1 < @x` (but in this trivial case `preselection='cs1 < %s' % x` would also have worked).
  

Out-of-core computations
---------------------------
What if you want to e.g. make a cs1/cs2 spectrum of all events? You can't use preselections to make the data fit into RAM. You could write a for loop over datasets, adding to the histogram on each pass, but that's tedious, and you'd have to write your own parallelization code to get it fast. Instead, try the delayed option:

    data = hax.minitrees.load(run_numbers, delayed=True)

which outputs the data as a [dask dataframe](http://dask.pydata.org/en/latest/dataframe.html) instead of a pandas dataframe. You can still use the hax.cuts functions on this, much of the pandas functionality, and make histograms using multihist (not plt.hist or np.histogram). You can't make scatter plots -- unless you use something like [datashader](https://github.com/bokeh/datashader), or reduce the data down to a manageable size. 

If you like more info on dask, see e.g. [this talk](https://www.youtube.com/watch?v=PAGjm4BMKlk), or several longer talks on youtube.

Unforunately cut history is currently not kept when doing out-of-core computations. This will be a bit of a challenge to implement in a nice way. 

Minor changes
---------------
* The use_pickle and use_root options have been removed. Instead you can now set the following hax.init options:
  * `preferred_minitree_format`. Format of minitrees that will be used for saving new minitrees and that will be searched for first.
  * `other_minitree_formats`. A list of other minitree formats to check for. We'll only check for these if we can't find any in your preferred format.
* This means you can no longer directly convert between minitree formats at the moment. You can still re-make a minitree in another format of course. We might want to re-enable this feature in the future.
* The pickle backend now uses compression (gzip).

### DIFF
--- a/examples/Preselections and out-of-core computation.ipynb
+++ b/examples/Preselections and out-of-core computation.ipynb
@@ -1,0 +1,1009 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Preselections and out-of-core computations with hax"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Jelle, 7 October 2016"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "/project/lgrandi/anaconda3/envs/pax_head/lib/python3.4/site-packages/IPython/kernel/__init__.py:13: ShimWarning: The `IPython.kernel` package has been deprecated. You should import from ipykernel or jupyter_client instead.\n",
+      "  \"You should import from ipykernel or jupyter_client instead.\", ShimWarning)\n"
+     ]
+    }
+   ],
+   "source": [
+    "import logging\n",
+    "logging.basicConfig(level=logging.INFO)\n",
+    "\n",
+    "import numpy as np\n",
+    "import pandas as pd\n",
+    "from tqdm import tqdm\n",
+    "import dask\n",
+    "\n",
+    "import matplotlib\n",
+    "%matplotlib inline\n",
+    "import matplotlib.pyplot as plt\n",
+    "\n",
+    "from pax.configuration import load_configuration\n",
+    "pax_config = load_configuration('XENON1T')\n",
+    "\n",
+    "import hax\n",
+    "from hax import cuts\n",
+    "# Load minitrees from several analysts kind enough to share their home directory\n",
+    "# If you think this is getting crazy, I agree... the solution is a common minitree folder\n",
+    "# with pre-made basics minitrees. But that would require someone maintaining it.\n",
+    "hax.init(minitree_paths=['.', \n",
+    "                         '/home/aalbers/minitrees', \n",
+    "                         '/home/breur/minitrees/v5.6.2', \n",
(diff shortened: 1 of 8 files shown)


tunnell:
Nice.  but Gah! I had the dask docs up this week .

JelleAalbers:
Haha, thanks :-) I used to think we'd be going with blaze, but dask seems more concrete now. The real kicker is dask.delayed, you can nicely transition between doing your own thing (but feeding into their task graph) and going with the full fake-pandas API they have.

If you're looking for a challenge though... the cuts history tracking gets quite complicated when doing out-of-core delayed processing. Not sure how that will even work without subclassing their dataframe object. Connecting with datashader or some other bokeh stuff would be nice too.

tunnell:
Playing with it.  There are some missing descriptions of what some variables do such as in load() for the number of workers.  Will give more comments affter playing.

tunnell:
OK, so I can add a config thing that then can reinitialize hax from the distributed job.  Working on it since it's a little tricky actually.

JelleAalbers:
I've worked with this for a few days and it seems to work fine. Any objections against merging this week?

tunnell:
That's fine.  I just wanted to have a cluster setup too, but this will require tweaking this a bit.  So it'll just become a new pull request.

PR #38, opened 2016-10-24
Small changes / fixes

JelleAalbers:
This includes a few small changes:

* Hax no longer tries to compile the pax C++ event class if you're loading cached minitrees. This used to happen when hax checked if the minitree you're trying to load had the right verion of pax -- but for that you only need to peek at the metadata stored in the root file, not look at actual events. This should make loading of premade minitrees faster and less error-prone.

* Fix a bug in cut history tracking where new dataframes could "magically" inherit the histories of old dataframes after you'd deleted them. (turns out the `id` function in python gives memory addresses, not unique hash-like values...).

* Add a new option `make_minitrees` to hax init, which is True by default. If you set it to False, hax will silently (well, with just a debug message) skip datasets for which it can't find an up-to-date minitree. Not something you want to use for a big final serious physics analysis, but quite useful during development / prototyping. If you know what you're doing.

### DIFF
--- a/hax/cuts.py
+++ b/hax/cuts.py
@@ -11,10 +11,6 @@ import hax
 import logging
 log = logging.getLogger('hax.cuts')
 
-# Dictionary mapping id of DataFrame objects to list of cut information applied to it.
-# Weakrefs would have been really nice here... but as DataFrames are mutable, they can't be hashed,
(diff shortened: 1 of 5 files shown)


JelleAalbers:
If there are no objections I will merge these tomorrow.

PR #40, opened 2016-11-15
Cache file, TotalPeakProperties

JelleAalbers:
This adds a TotalPeakProperties treemaker which provides:
   - total_peak_area, the total area (pe) in all peaks in the event (even non-tpc peaks)
   - n_pulses, the total number of pulses in the event (for pax versions >6.0.0
   - n_peaks, the total number of peaks in the event (including non-tpc peaks and lone hits)

Also, this adds a cache_file option for minitrees.load:
  * If you load many minitrees, the load procedure can take more than a minute (even if the minitrees have already been made, even if you use several cores with num_workers). 
  * If you specify cache_file you can save the result of the load + preselections to a local file.
  * The next time you load the data it will take just a second or so.

You can copy this file to another computer and do analysis with it. You can also use it to keep your analysis stable even if minitrees get removed or updated. To remake the cache file, just remove the file, or pass remake_cache=True. Don't confuse remake_cache with force_reload=True, which will remake all the minitrees. Do NOT use this if you're loading in a lot of data without strong preselections!


JelleAalbers:
If there are no objections I will merge this tomorrow.

PR #54, opened 2016-12-16
Blinding proposal

tunnell:
Checks if data is marked as blinded in the run db and applies blinding cut if so.  If can't find information on dataset, blinds by default.    More information in https://xecluster.lngs.infn.it/dokuwiki/doku.php?id=xenon:xenon1t:analysis:tunnell:tempblinding

### DIFF
--- a/hax/hax.ini
+++ b/hax/hax.ini
@@ -87,7 +87,7 @@ raw_data_local_path = '.'
 ##
 # Blinding cut
 ##
-cut = '(cs1 > 15) & ((pid > exp(-0.989158 + (-0.026168) * cs1) + 1.918830 + (-5.722440e-04) * cs1) | (cs1 > 200))'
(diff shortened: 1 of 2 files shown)


hax/hax.ini
```diff
@@ -84,6 +84,10 @@ main_data_paths = ['.', './data']
 
 raw_data_local_path = '.'
 
+##
+# Blinding cut
+##
+cut = '(cs1 > 15) & ((pid > exp(-0.989158 + (-0.026168) * cs1) + 1.918830 + (-5.722440e-04) * cs1) | (cs1 > 200))'
```
JelleAalbers: Could you name it blinding_cut?

hax/minitrees.py
```diff
@@ -326,6 +351,10 @@ def load_single_dataset(run_id, treemakers, preselection=None, force_reload=Fals
     for i in range(1, len(dataframes)):
         result = _merge_minitrees(result, dataframes[i])
 
+    #
+    if is_blind(run_id):
```
JelleAalbers: This will error if you're not loading basics, since the variables in the blinding cut are not defined

tunnell:
Implemented change @JelleAalbers

PR #61, opened 2017-01-09
Apply blinding cut to all datasets if one dataset requires it

JelleAalbers:
Currently, if you attempt to load both blinded and unblinded data, hax crashes. This is because the cuts history tracking assumes the same cuts are applied to all datasets you're loading in, but the blinding cut gets applied at the per-dataset level (in hax.minitrees.load_single_dataset, not hax.minitrees.load).

This adds extra logic to hax.minitrees.load to ensure that, if any dataset you selected requires the blinding cut (and you have basics in the minitrees you're loading) the blinding cut is applied to all of them. I tested it briefly and it seems to work. 

Thanks @sanderbreur for reporting the issue. @tunnell could you have a brief look, since it affects the blinding code you wrote?


JelleAalbers:
Thanks!

PR #63, opened 2017-01-11
Small update to minitree with arrays example

jhowl01:
Updates the previous example usage of minitrees containing array branches to incorporate more recent changes in the way hax saves minitrees:

- Demonstrates the `uses_arrays` option.

- Rather than saving information on all peaks (an application for which the MultipleRowExtractor is more effective), the example now just looks at the largest five. 

- Got rid of outdated `use_root` and `use_pickle` options, which have been replaced by `preferred_minitree_format`.


JelleAalbers:
Thanks Joey!

PR #64, opened 2017-01-24
Update slow control interface

JelleAalbers:
The current hax slow control interface is not working, since it relies on a database server that has  disappeared. This adds a new interface that works in a similar way, but uses the slow control group's HTTP API (https://xecluster.lngs.infn.it/dokuwiki/doku.php?id=xenon:xenon1t:slowcontrol:webservice). 

To get the cathode voltages measured in run 6000, you would do: 
```
data = hax.slow_control.get('cathode', 6000)
```
Now `data` is a pandas series with UTC time on the index; you can plot it, do things like `data.mean()`, etc.

* You don't have to specify the historian ID of the variable you want; hax does some very basic fuzzy matching so you only have to specify part of the description (comment) of the variable. If what you ask for is known uniquely, you get it, if not you get an exception listing the known alternatives. 
* You can also use historian ids ('XE1T.GEN_HEINZVMON.PI'), slow control ids ('TPC_Monitor_Voltage'), or "Pid identifiers" (none for the cathode, but e.g. "SLM01" for short level meter 1). Here no fuzzy matching is done, you have to be exact.

Instead of a run number you can specify a particular time range. For example, to get and plot the PT102 values since yesterday noon:
```
data = hax.slow_control.get('PT102', start='noon yesterday', end='now')
plt.plot(data)
plt.gcf().autofmt_xdate()
```
pt102

We use the very nice library https://pypi.python.org/pypi/parsedatetime/ to parse the start and stop datetimes, so you can specify them in pretty much any format you want. 

As always it's possible there is an error in the timezones (though I tried to be careful and UTC throughout). Please let me know if this is the case.

If you want a nice slow control viewer in your notebook, have a look at https://github.com/XENON1T/XeAnalysisScripts/tree/a40e02a7ab9b8d6ee6e8549817e93f7964db5119/SlowControl/HistorianReader from @acolijn. I took the file with slow control aliases from there (and added the PMT/anode high voltage variables).


tunnell:
Nice!  Can you still fetch a list of SC variables since you took it out of the slow_control.py?  That was meant to make it programmatically easier to loop over different classes of variables. Also, why encrypt the API key?  Can't we just set that in the environment like the mongo password?

tunnell:
I think people should use their own API key.  Otherwise, we can't revoke abusive scripts.

JelleAalbers:
Thanks for having a look so quickly. I've added support for asking for multiple slow control parameters in one go back in now, so you can do for example:
```
df = hax.slow_control.get(['PT102', 'cathode'], start='noon yesterday', end='8pm')
```
and get a dataframe with 'PT102' and 'cathode' columns (and time on the index as usual). 

Putting the API key in the environment would also have been a solution, but that would have required me to adjust the deploy scripts / the environment setup scripts; I think this is at least equally simple. Having separate API keys would be useful if people start using this heavily, though we'd have to think of a cleverer distribution system than than letting everyone email Miguel Silva. At the moment I think having this 'batteries included' is best. The SC group already set up some throttling to prevent over-use of the API (https://xecluster.lngs.infn.it/dokuwiki/doku.php?id=xenon:xenon1t:slowcontrol:webservice).

tunnell:
If you make a variable for SC API key, then I'll add it to the deployment then merge.  It's the right way to do it.  This way, the batteries are included but we give ourself freedom in future.  Also, this Mongo password I keep meaning to change to it'll be super annoying to change one thing and break another.

JelleAalbers:
Ok, good point, we do want to change the mongo password. I configured it to look for the SC_API_KEY variable. If necessary you can override in a script by manually passing sc_api_key to hax.init (just like mongo_password).

tunnell:
Can you email me the API key to set on Midway?

tunnell:
set in env

pdeperio:
Getting this error running ```haxer``` from CVMFS (pax_v6.2.1):
```
FileNotFoundError: File b'/cvmfs/xenon.opensciencegrid.org/releases/anaconda/2.4/envs/pax_v6.2.1/lib/python3.4/site-packages/hax-1.3.0-py3.4.egg/hax/sc_variables.csv' does not exist
```
Seems like ```python setup.py install``` is not copying the file?

pdeperio:
I just copied the file manually for now:
```
/cvmfs/xenon.opensciencegrid.org/releases/anaconda/2.4/envs/pax_[head, v6.2.1]/lib/python3.4/site-packages/hax-1.3.0-py3.4.egg/hax/sc_variables.csv
```
and
```
/project/lgrandi/anaconda3/envs/pax_[head, v6.2.0, v6.2.1]/lib/python3.4/site-packages/hax-1.3.0-py3.4.egg/hax/sc_variables.csv
```
but should figure out why deploy failed.

JelleAalbers:
My fault, sorry: https://github.com/XENON1T/hax/issues/66. Thanks for fixing it manually.

PR #68, opened 2017-02-09
Changed raw data handling to allow for multiple locations in local mode

DCichon:
The changes should allow to handle multiple local raw data locations. Quick tests with a script which uses hax's raw data interface did not show any problems or undesired behavior. If possible and/or necessary, additional tests for trigger data access could be done using a corresponding analysis script, but one would need to point me towards it.


JelleAalbers:
Looks very nice Dominic, thanks! I'll be away for the next few days, but I'll have a look at it once I'm in Chicago. Just from a quick glance, it might be nice to also allow people to specify a single path (e.g. by 
```
if isinstance(hax.config['raw_data_local_path'], str):
    hax.config['raw_data_local_path'] = [hax.config['raw_data_local_path']]
```
in hax.init (in `__init__.py`). That would enable the current code that uses this to keep working.

DCichon:
Thanks for your suggestions, Jelle. They have been implemented along with some minor additional refinements.

pdeperio:
Would it be possible to also read a directory containing a single .zip file as long as a valid event number is specified? E.g.:
```
/project/lgrandi/xenon1t/raw_for_waveforms/170209_1703
```

I quickly tried in master and get the error:
```
/project/lgrandi/anaconda3/envs/pax_head/lib/python3.4/site-packages/hax-1.3.0-py3.4.egg/hax/raw_data.py in process_events(run_id, event_numbers, config_override)
    118         # We can let pax deal with jumping from file to file, selecting events, etc.
    119         if not dataset_info.raw_data_found:
--> 120             raise ValueError("Raw data for dataset %d (%s) not found." % (run_number, run_name))
    121         dirname = os.path.join(config['raw_data_local_path'],
    122                                dataset_info.raw_data_subfolder,

ValueError: Raw data for dataset 6959 (170209_1703) not found.
```

DCichon:
Does it work when you set "raw_data_local_path" to "/project/lgrandi/xenon1t/raw_for_waveforms"? Hax looks for folders matching the run folder name, but not for single .zip files.

pdeperio:
sorry, my mistake. indeed ```/project/lgrandi/xenon1t/raw_for_waveforms``` works!

PR #83, opened 2017-03-02
Updated proximity trees with new branches

coderdj:
Adds the following:

previous_busy_on
next_busy_on
nearest_busy_on

previous_hev_on
next_hev_on
nearest_hev_on

previous_s2_area
next_s2_area
nearest_s2_area

The former are to include in the DAQ cut for the 'sub-prime' period. The latter is for the self-veto cut.

### DIFF
--- a/hax/treemakers/trigger.py
+++ b/hax/treemakers/trigger.py
@@ -135,3 +135,30 @@ class Proximity(hax.minitrees.TreeMaker):
             result['nearest_s2_area'] = result['next_s2_area']
 
         return result
+
+class TailCut(hax.minitrees.TreeMaker):
+
+    __version__ = '0.0.2'
+
+    def get_data(self, dataset, event_list=None):
+
+        self.event_data = hax.minitrees.load_single_dataset(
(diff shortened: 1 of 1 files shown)


JelleAalbers:
Thanks! Why include the s2 area? That's already in basics.

If we merge this all existing Proximity minitrees will be invalidated for everyone, and they'll have to be remade. It may be good to merge this at an inconspicous time, then start jobs to remake Proximity for all processed 6.4.2 data.

coderdj:
I thought it might help for preselections to have the s2_area stored in the event even if the previous event is rejected.

JelleAalbers:
Ah, sorry, you included the s2 area of the previous and nearest event. Yes, that makes sense, sorry.

hax/treemakers/trigger.py
```diff
@@ -114,4 +128,37 @@ def extract_data(self, event):
...
+
         return result
+
+class TailCut(hax.minitrees.TreeMaker):
+
+    __version__ = '0.0.2'
+
+    def get_data(self, dataset, event_list=None):
+
+        self.event_data = hax.minitrees.load_single_dataset(
```
tunnell: remove line break

PR #84, opened 2017-03-03
Added XENON1T/hax/hax/treemakers/MuonVetoTreemaker.py

geisch:
Added the `MuonVetoTreemaker`, where, so far, all muon veto based analysis is based on. Adding it on the main hax repository and thus making it able to be loaded with the `hax.minitrees.load()` function will make it easier to use and easier to adapt and improve in future as well.

A check if it is found by hax automatically in this way has to be done


pdeperio:
accidentally merged somehow...

PR #91, opened 2017-03-10
Revert "Added XENON1T/hax/hax/treemakers/MuonVetoTreemaker.py"

pdeperio:
Reverts XENON1T/hax#84


JelleAalbers:
Thanks for reverting, but after #92 I presume this doesn't hurt anyone. I'll speak with Christopher about any suggestions for the muon veto treemaker.

PR #96, opened 2017-03-14
added a variable check in the last for loop of get_sc_data function

lucrlom:
It's necessary to verify that the r.json provides a dictionary type variable `dict`. If some issue with the slow control data base connection happen or the data are missed, the `entry` variable is no more a dict and an error occur when try to read the `entry['timestampseconds']` data.
I'm not sure if put `continue` is the best solution or some alternative action can be implemented.


JelleAalbers:
Thanks Francesco! I didn't know the API could return things other than dicts; clearly we can't parse them though, so this will help.

PR #98, opened 2017-04-04
unblind multiple scatters

feigaodm:
This change is to unblind multiple scatter events in the dark matter search data. The motivation for this change is to investigate poteintial multiple scatter neutron background in the detector in order to understand the single scatter neutron background better. An example analysis can be found in my [earlier note](https://xecluster.lngs.infn.it/dokuwiki/lib/exe/fetch.php?media=xenon:xenon1t:analysis:double_scatter_postblinding_0322.html)


tunnell:
I think you need to make the double scatter requirement less weak.  Right now it's 150, but at what energy does that correspond to unblinding since the S2[1] scales as S2[0]?  Can you do 200 or higher so we know we're never analyzing these energies?

feigaodm:
So you are worrying that we will stick to the same S2 threshold of 200PE for SR1 analysis. I can make the change to that. Accoring to [Tianyu's latest single scatter cut using pax_6.4.2], the threshold is s2 0.01*s2 + 90.0. So 200PE threshold corresponds to a main S2 of 1.1e4 and should be sufficiency for our signal region of interest (probably below 100PE s1).

PR #99, opened 2017-04-06
unblind multiple scatters

feigaodm:
This change is to unblind multiple scatter events in the dark matter search data. The motivation for this change is to investigate poteintial multiple scatter neutron background in the detector in order to understand the single scatter neutron background better. An example analysis can be found in my [earlier note](https://xecluster.lngs.infn.it/dokuwiki/lib/exe/fetch.php?media=xenon:xenon1t:analysis:double_scatter_postblinding_0322.html). Threshold is updated to 200PE in uncorrected-S2.


feigaodm:
@tunnell Should we unblind S2 below 200PE as well? Or we keep it as 150PE.

feigaodm:
@tunnell Should we merge this now or later? I think it would be better to make it earlier to make more time for the multiple scatter analysis in SR1

PR #100, opened 2017-05-10
Better blinding

coderdj:
Adds parameter 'blind_from_run'. Setting this to a run number automatically blinds *all* background_stable runs after this point unless they are specifically tagged 'unblinded' or '_unblinded'. 

This should make it so shifters don't have to put the blinded tag on each run and should remove the possibility of accidentally unblinding because a shifter forgets or misspells this tag.


JelleAalbers:
Thanks Dan, this looks fine. I took the liberty of adding a few comments, fixing a typo ('number' -> number), and removing an unneeded except (KeyError should happen when using get_run_info asking for a single field). If you agree feel free to merge (unless @feigaodm sees a problem?)

coderdj:
I have no problem with merging but I'd leave the #blinded tag in until at least a couple people try this out on a few (unblinded) runs. I'll wait for advice from Fei though.

feigaodm:
Sorry that I responded to @coderdj only, definitely it's a better way for blinding. Maybe we can test it with new runs without 'blinded' tag, but it's not a stopper to to merge it

PR #101, opened 2017-05-11
Add Adam's S2TopBottom minitree

JelleAalbers:
This add's Adam's S2TopBottom minitree for sciencerun0 to hax. This tree was created to store updated S2 corrections for science run 0. For future runs this specific treemaker should probably be updated (or cs2_bottom should included in some other treemaker), but it's used in many current analyses, for which we are now copying adam's file around.


feigaodm:
Thanks Jelle, How about I copy your code here to extended minitrees? I think we want to avoid too many minitrees in SR1 analysis, that's why you collect the needed branches in "extended" minitree.

JelleAalbers:
Hi Fei, that's a good idea in principe, as you say we don't want too many minitrees. But if we update an existing minitree, all current minitrees will be invalidated (hax won't load old versions of minitrees). We don't really have a workflow for re-making minitrees yet, so this would require some manual work. If you're willing to (find someone to) do that it is certainly the best solution, but it will be nontrivial -- considering also older pax versions, some of which may not even have their processed data available anymore.

(p.s. this is really Adam's code ;-) )

coderdj:
What would you think about a single minitree for all post-processing corrections? Then we'd just have one thing to recompute in case corrections change.

feigaodm:
I agree with Dan to have a signal correction minitrees for all the post-processing corrections. I can start with Jelle/Adam's minitree and keep you updated.

coderdj:
@feigaodm I started a bit here: https://github.com/XENON1T/hax/tree/s2_xy_haxit
Don't know if that's helpful. Some of it is reproducing what Jelle did, some is pulling from DB. I thought to add caching from the DB (basically pulling corrections docs at startup only if they don't exist in cache already) in order to cut down on download sizes, since some of these are pretty big. 

I don't have time to do more on this this week but can work again on it next week. We should just mess around a bit and figure out what works best. I was also debating if putting the corrections in DB docs is reasonable (they get bit) or if we should just serve full files.

adambrown1:
How about we rename this tree to "Corrections" and add it now

adambrown1:
Ho

feigaodm:
Should be covered by PR #109

PR #105, opened 2017-05-30
Add LNGS as analysis site for backup

pdeperio:
Add LNGS (xecluster, ULITE) spoofed as ```xe1t-datamanager```


pdeperio:
actually wait, the ```processed``` and ```minitree``` locations need to be updated @rizzoa

PR #106, opened 2017-06-01
Add replica MongoDBs

pdeperio:
for more redundancy if Stockholm DB goes down


pdeperio:
> Daniel Coderre @coderdj 09:36
> @pdeperio is there a way for pymongo to set "SecondaryPreferred"? Then we could use the replica set
>
>Patrick de Perio @pdeperio 09:37
>dunno
>
>Daniel Coderre @coderdj 09:37
>There is. So for now I'd say just put the stockholm mirror for hax
>In the future we could consider putting the whole set with 'secondary_preferred' set to true so that zenigata gets used too. Only thing is analysis would only work at sites whitelisted by all 3 places

PR #107, opened 2017-06-06
Minitree bugfix

### LINKED ISSUE #47: Missing Branches in Output
The ```lone_hit_*``` branches are missing in some ```LargestPeakProperties``` minitrees, causing problems in the MC workflow. Can reproduce with files on Midway in:
```
/project/lgrandi/pdeperio/161206-hax_debug
```
by running:
```
source activate pax_v6.1.1
HAXPYTHON="import hax; "
HAXPYTHON+="hax.init(main_data_paths=['/project/lgrandi/pdeperio/161206-hax_debug'], minitree_paths=['.'], pax_version_policy = 'loose'); "
HAXPYTHON+="hax.minitrees.load('Xenon1T_TPC_Kr83m_00000_g4mc_NEST_Patch_pax', ['LargestPeakProperties']);"; python -c "${HAXPYTHON}" # Seems OK
HAXPYTHON+="hax.minitrees.load('Xenon1T_TPC_Kr83m_00183_g4mc_NEST_Patch_pax', ['LargestPeakProperties']);"; python -c "${HAXPYTHON}" # Missing branches
```
producing the two files:
```
Xenon1T_TPC_Kr83m_00000_g4mc_NEST_Patch_pax_LargestPeakProperties.root # Seems OK
Xenon1T_TPC_Kr83m_00183_g4mc_NEST_Patch_pax_LargestPeakProperties.root # Missing branches
```

feigaodm:
* This PR should (fix #47) 
  * The whole hax code has been cleaned to avoid style issues


JelleAalbers:
Thanks for the fix. 

The style change is at least somewhat a matter of preference, and my preference is obviously exactly the way it was, so you should not ask me :-) But e.g. this
https://github.com/XENON1T/hax/pull/107/files#diff-8c7bfe760fb013f7ec520a663b1a4ddbR248
doesn't look like a change for the better. 

Did you go over all this yourself or ran an autoformat tool? If the latter (which I have very bad experiences with) it might at least be worth redoing with a different maximum line length. In hax, pax, etc. we always worked with a 120-character limit (in hax soft limit); reformatting things to e.g. 80 characters automatically probably does not make it better, unless you're prepared to extract variables, break up list comprehensions, etc.

You may like this talk: https://www.youtube.com/watch?v=wf-BqAjZb8M

feigaodm:
Thanks @JelleAalbers .  I was just trying to avoid the fact that make lint fails for hax:) Most of those style issue appears in the TreeMakers random person created. I use a software to correct it automatically. After this fix, I tested it to make sure it works. 

I will take a look at your suggested link sometime.

PR #108, opened 2017-06-09
added automatic printing of tag version if there is a version

coderdj:
If someone uses hax to select data with a tag, and if that tag has a corresponding 'version' then we print the version and date of the tag to std out. Combined with the #sciencerunx versioning database this will allow us to determine exactly which good run list is used for a given analysis.


coderdj:
Note: this was a web-browser-editor commit

feigaodm:
@coderdj Thanks for this PR, I probably merged it too fast. I saw a potential problem: 
at Line 344:16, is  'collection' defined somewhere?

PR #112, opened 2017-06-13
Backwards compatibility for s1_area_fraction_top_probability

darrylmasson:
Should prevent failures when growing minitrees earlier than pax_v6.6.0. In these cases the `s1_area_fraction_top_probability` isn't stored in the minitree.


feigaodm:
Thanks for this PR, but as Jelle said, the version bump may force reloading the minitree again, can you put the version back to 0.03 for now?

coderdj:
Bump the version! We want to regrow minitrees since this cut is now required for SR1.

PR #113, opened 2017-06-14
module import bug fix

feigaodm:
This should fix the module import error various analyzers met. Thanks @JelleAalbers and @pdeperio to check this.


hax/__init__.py
```diff
@@ -3,9 +3,8 @@
 import inspect
 from configparser import ConfigParser
 import socket
-# from . import misc, minitrees, paxroot, pmt_plot, raw_data, runs, utils, treemakers, data_extractor, slow_control, \
-#    trigger_data, ipython, recorrect
-from . import runs, minitrees, slow_control
+from . import misc, minitrees, paxroot, pmt_plot, raw_data, runs, utils, treemakers, data_extractor, slow_control, \
```
pdeperio: Move the "stitching" comment below back to above this line

PR #114, opened 2017-06-20
Previous event Basics information

JelleAalbers:
This adds the PreviousEventBasics treemaker, which brings you the info from Basics for the previous event. It's main use case is getting information about the directly preceding event when you are doing analyses with some cuts or preselections (so you can't just look at the previous row in the dataframe you have left).

This treemaker doesn't need to write its own minitree files: it just loads Basics and shifts the data by one. To support this I added a `never_store` boolean attribute for minitrees. If you set this to True (default is False), no minitree files will be created for a treemaker. This could have broader use, e.g. in performing some quick distributed computation during the minitree loading process.


coderdj:
What about blinding? Looking through the code it looks like this might be an issue.

JelleAalbers:
Nice, hadn't considered this. As long as you include Basics in the list of trees you load, you are fine. If not you don't get blinding. So you can already plot largest_s1 and s2 from TotalPeakProperties, or even better the int_a stuff from DoubleScatter, and accidentally unblind...

Maybe we should mandate the inclusion of Basics (and therefore the blinding cut) if a blinded run is loaded? I thought we didn't do this for some reason (which I can't remember now).

coderdj:
Are you sure it's fine to just include basics? I don't think blinded events usually follow each other. So my understanding is including basics would correctly blind the blinded events, but if subsequent events are not blinded then the blinded events would still be available with this tree. 

(Reading this sentence a few times not sure if it's clear)

Maybe we need to add blinding logic to this tree. If the event is blinded set everything to '0' and add a row 'previous_event_blinded'?

JelleAalbers:
Yes, this minitree is a bit special, sorry. The risk here is different than for DoubleScatter, since it contains information about different events. 

However, if you just plot previous_cs1 vs previous_cs2 for all events without preselections (if you can even manage to do this with the massive data volume it requires) you won't get something useful: see e.g. figure 1 here: 
https://xecluster.lngs.infn.it/dokuwiki/doku.php?id=xenon:xenon1t:aalbers:2days_neverblinded

There are lots of high-r events, cathode events, surface events, gas events, and junk events masking anything worth knowing at low energy. Only when you start applying cuts do you get something interesting. So you'd have to modify all cuts to act on previous_x to circumvent blinding with this tree -- but if you _want_ to circumvent blinding, it's much easier to just set the blinding cut to something else in hax.init (like we do for MC data). The key is to prevent someone unblinding by mistake, this tree doesn't make that any more likely I think.

coderdj:
What if we just didn't include cs1 and cs2 in the previous minitree? I can't think of a situation where this would be needed, as opposed to uncorrected values, which I can imagine.

tunnell:
We still want this, right?

coderdj:
Did we determine for sure it doesn't accidentally unblind the data?

coderdj:
Plot previous event cs1 vs cs2 and you get it even if previous event blinded, right?

JelleAalbers:
Deleting cSx will only help a bit, the ER/NR difference is larger than our corrections.

The main point is that without cuts you [learn nothing](https://xecluster.lngs.infn.it/dokuwiki/doku.php?id=xenon:xenon1t:aalbers:2days_neverblinded#looking_at_data). You could hack some cuts to work on previous_X, but there are easier ways to circumvent blinding (just disable blinding cut).

I'm fine with this not going into hax if people see issues though, I rarely use the previous_X stuff and when needed I can just copy the class into my notebook. If we want to go this way I'll make a separate pull request for the never_store option, which an be used for other minitrees (e.g. TailCut).

feigaodm:
Will we define so many cuts to analyze previous event's information and search for WIMPs using it? Otherwise we won't see a problem with this treemaker. Especially since @coderdj removed cs1 and cs2 from Basics minitree, we should not have any problem, right. In addition, a lot of cuts are defined based on other treemaker which we don't have for previous event, so I don't think it will affect the blinding. Just warn people not to dig too much in this previous event properties. 

In short, I'm fine with it, and maybe it's helpful to have this for some interesting studies, like S2-only? Maybe @JelleAalbers has better motivations than this.

PR #118, opened 2017-06-30
Extended minitree bug

feigaodm:
This PR is triggered by an issue seen from a full chain simulation/analysis MC->fax->pax->hax->lax. When we apply S1 single scatter cuts, we used the "alt_s1_interaction_z" variables to see this S1 is consistent with some S2s. However, there are some simulated events which don't have any other S1s than the real S1s, and thus "alt_s1_interaction_z" does not exist in the mini-trees. 

In this PR, we initialised the alternative variables to be float("nan") so that people still can access it to define various cuts. As a result, people defining cuts on other S1/S2 peaks should check the effect on their cuts, like S1SingleScatter @jpienaar13 and S2SingleScatter @zhut19 cuts.

Alternatively, people working on those cuts should check whether the branches they used are stored inside the dataframe. The fact that those variable doesn't exist can already be used to judge whether that event should pass or not.


jpienaar13:
This was an issue considered when creating this cut. IT should alreay return a true value when encountering a nan for the alt_s1_interaction_z

feigaodm:
@jpienaar13 Yeah, but you guys should consider modifying your cuts in LAX instead:)

jpienaar13:
Wait, I'm confused, do we not want the cut in LAX to give a true value if no second S1 exists?

feigaodm:
Yes, we want it. My warning is not only to you, but also to others like @zhut19 who defines the cut on those variables.

jpienaar13:
Ah ok. Thanks for the clarification.

PR #119, opened 2017-07-05
Electron lifetime correction to minitree

coderdj:
**Don't merge this yet:** We need the updated electron lifetime function from Zach in the DB before it makes sense to use. But I wanted to put it here so it's on the roadmap.

This commit expands Adam's corrections minitree to add the electron lifetime correction. The electron lifetime correction is defined as a list of (timestamp, lifetime) values pulled from MongoDB. We kept MongoDB instead of moving to a calibration file because we expect this correction to be updated very often during a science run and it's easier to update via database.

The cs2, cs2_top, and cs2_bottom values are all corrected with this correction. Additionally it is stored in the corrections minitree such that the original lifetime can be recalculated using the correction value and the drift time.

The correction is based on the _run_ start time, not the event time. Appropriate access functions are added. Also a method for fetching corrections is added to runs.py and run on hax.init.

I tested this on a few runs and it at least doesn't crash and gives values close to pax's cs2. The small differences could come from either using wrong lifetime values (I'm using a file I got from Pietro's home directory of unknown origin) or interference from other corrections.


coderdj:
OK I now made several updates and I think this is about ready for prime time. The following corrections are being applied:
  * Electron lifetime (from DB)
  * S2 xy map (file)
  * S1 LCE maps (file)
  * FDC simple version (file)

In addition now the *only* place to get cs1 and cs2 is the corrections minitree. They have been removed from the basics minitree and the basics minitree version was bumped up. In addition the blinding logic now asks if the corrections (not basics) minitree has been loaded before applying the blinding cut.

This could create a tricky blinding situation so I think the best thing to do would probably be to manually purge all old 'basics' minitrees for all pax versions once this is merged and regenerate them. 

Comments welcome.

hax/treemakers/corrections.py
```diff
@@ -87,4 +168,19 @@ def extract_data(self, event):
         result['cs2_top'] = s2.area * s2.area_fraction_top * s2_top_correction
         result['cs2_bottom'] = s2.area * (1.0 - s2.area_fraction_top) * s2_bottom_correction
```
feigaodm: where is cs2 variable after this change? If we don't have cs2, the blinding cut will fail

coderdj: cs2 is in the corrections minitree only. Blinding is changed so that it only runs if corrections loaded.

I'm not entirely sure this is the best but we don't want cs2 in basics. It will not be the same as cs2 from corrections so this is too confusing.

coderdj:
Newest update fine tunes blinding behavior. This is a bit critical so I'd like to get an opinon.

 * cs1 and cs2 removed from Basics
 * blinding logic only done if 'corrections' loaded. We will no longer blind if Basics is loaded only (but since Basics doesn't contain cs1 or cs2 should be ok)
 * As a slightly nasty workaround the variables 's2' and 'largest_other_s2' are added to the corrections minitree. I don't like having the same variable defined in two places but couldn't think of another way to have these variables there for the blinding cut.

hax/treemakers/corrections.py
feigaodm: My question is I can not find a specific line that defines 'cs2' in the tree maker. For eg, I see in L168 'cs2_top' is defined, can you point to me where is 'cs2'

feigaodm: never mind, I found them.

PR #121, opened 2017-07-11
Add three more items to Extended for S2Single cut

zhut19:
Add properties of largest other s2 to Extended tree maker, which are used in sr1 S2 single scatter cut. Called largest_other_s2_delay_s1, largest_other_s2_delay_s2 and largest_other_s2_pattern_fit respectively.


hax/treemakers/common.py
```diff
@@ -126,6 +130,22 @@ def extract_data(self, event):
...
 
+        result['largest_other_s2_delay_s1'] = float('nan')
+        result['largest_other_s2_delay_s2'] = float('nan')
+        result['largest_other_s2_pattern_fit'] = float('nan')
+
+        for ix in event.s2s:
+            if (ix != interaction.s2) and (event.peaks[ix].hit_time_mean - s1.hit_time_mean > 0):
+                # largest other s2 after main s1
+                pk = event.peaks[ix]
+                result['largest_other_s2_delay_s1'] = pk.hit_time_mean - s1.hit_time_mean
```
feigaodm: It looks like this is not necessary since we have a branch called alt_s2_interaction_z.

zhut19: There is a slight difference, that alt_s2_interaction will be 'nan' if largest other s2 happen before s1, while in my code, the largest among all s2s after main s1 is selected. 

Maybe using largest_other_s2 in the whole event is better. But still couldn't just use alt_s2_interaction here, because it return 'nan' sometimes.

feigaodm: ok, thanks for clarification

PR #125, opened 2017-07-21
Proximity class modification

mlotfiben:
I have add modification to the Proximity inline with what is needed to calculate the nearest MV trigger to the TPC trigger.
In these changes I have included two new functions and in addition a time correction to take into consideration the reference of the TPC trigger time.
The two functions added do the following:

1.  In the MV data we have a new stream that is the MV synchronisation signal that is read by the TPC as a physical trigger sent by the MV. Which is in reality not true, because this signal was included just to keep the rollover of the digitiser working properly due to the low MV trigger rate and to be able to synchronise the two detectors. This signal has to be taken away from the acqm_monitor files to be able to have only the real physical MV triggers.

2. Also in the MV data stream sent to the TPC, afterpulses are still present and are taken as a physical triggers to calculate the coincidence between the TPC and the MV. These afterpulses are cleaned away.


tunnell:
@mlotfiben Please see why tests failing (Codacy).  You can click "Details" above.

mlotfiben:
@tunnell thanks, the tool is very useful, now everything seem to be fine.

PR #127, opened 2017-08-08
Updated TailCut in trigger.py

ramirezdiego:
Updated TailCut to a faster version, 0.0.3, initially improved by Jelle. This new version also saves the s2_area of the peak defining s2_over_tdiff, which might help to asses the quality of the cut on different values of this parameter.


JelleAalbers:
The never_store feature is currently in #114 which is still under review -- though it's mostly just the one line https://github.com/XENON1T/hax/pull/114/files#diff-c1f9d38420761a1ce52785cd25c65968R340 (and setting never_store=False by default).

PR #140, opened 2017-08-29
Fix S2 map version

pdeperio:
According to https://github.com/XENON1T/pax/pull/570 and the [pax XENON1T.ini](https://github.com/XENON1T/pax/blob/v6.8.0/pax/config/XENON1T.ini#L266), the S2 map should be ```v2.1```, not ```v2.0```.  Will need to regrow minitrees after merging this.


pdeperio:
@coderdj can you review and merge?

coderdj:
I think this change gets overwritten by a later PR but go ahead and merge this first. I think Adam has another s2 map in the works with the Kr electron lifetime correction so maybe should wait for that before regrowing.

PR #141, opened 2017-09-14
Added corrections definitions to minitree metadata

coderdj:
There's now a new field in minitree metadata called 'extra'. This field is an empty dict by default. It allows trees to set any additional information in metadata that may be pertinent. 

In principle this should rarely be used and the version should be used to keep things straight. However in the case of the corrections minitrees, the electron lifetime correction comes from a database and can therefore change without the tree version getting bumped. So we need some way to track that. This is the best I could come up with. I'm preparing a validation script to check for the presence and proper versioning of all official minitrees and this field will be part of that script.

Also added some more general code for the file-based corrections and stored them in the metadata as well.


hax/hax.ini
```diff
@@ -110,6 +110,25 @@ sc_variable_list = hax_dir + '/sc_variables.csv'
...
 
+##
+# Corrections Definitions
+##
+# The the electron lifetime is pulled from a remote database
+# File-based corrections defined here. Run ranges are inclusive
+corrections_definitions = {
+    "s2_xy_map": [
+         {"run_min": 0, "run_max": 6385, "file_name": "s2_xy_XENON1T_24Feb2017.json"},
+	 {"run_min": 6386, "file_name": "s2_xy_map_v2.0.json"}
```
pdeperio: v2.1

PR #142, opened 2017-09-14
Skip strings in acquisition monitor files (see pax#607 and pax#608)

JelleAalbers:
This makes the acquisition monitor pulse file reading code robust to the change introduced in https://github.com/XENON1T/pax/pull/608 . Files from old runs will remain readable.


JelleAalbers:
You could replace them with a random pickle too if they're causing problems -- I can only change what happens for new runs. I think an error is useful if the file is missing; it could be missing for many reasons.

XeBoris:
@JelleAalbers It sounds not useful to replace zero-sized *pickle files in the rucio catalogue by some file which has some minor random content. I know only of @ershockley who downloads this file from the rucio catalogue and put them into a folder for the analysts. Does hax itself need the file?

JelleAalbers:
Evan downloads them because they're needed for making the proximity minitrees. There's a difference between an absent acquisition monitor file (indicating some transfer issue) or an empty/header-only file (indicating absence of aq.m. signals, likely a DAQ problem). I think we had both kinds of issues in the past. 

If I'd have known Rucio doesn't like empty files I wouldn't have made them, but preventing their creation in the future is all I can do for you now.

Issue #139, opened 2017-08-24
Extending Corrections Minitrees to Work with MC simulations


jpienaar13:
Update to hax v1.6.2 caused problems in the production of minitrees for MC studies. (See #135). This has been temporarily circumvented by reverting to the last working hax commit (aadb15b) CVMFS.

Moving forward we will need an implementation  of a config option for all run dependent parameters that allows input for MC simulations to be set.

pdeperio:
Any decision yet on how to move forward with this?  Issue https://github.com/XENON1T/hax/issues/135 is technically still open, i.e. hax still not synced between data (Midway) and MC (CVMFS), and we should get this fixed ASAP since ```v6.8.0``` reprocessing is finishing.

I tried to start something coarse (science run variation) and basic in https://github.com/XENON1T/pax/pull/610, where you can use ```event.dataset_name``` (containing the string "MC") as a flag to determine what to do in [corrections.py](https://github.com/XENON1T/hax/blob/master/hax/treemakers/corrections.py).  But then here can you read in those new pax configs (```SR0_parameters.ini``` or ```SR1_parameters.ini```) and use it to set the e-lifetime (```elife_interpolation```) and all the correction maps (```get_lce_map_name ```: ```s1_light_yield_map```, ```get_s2_map_name```: ```s2_light_yield_map```, ```get_fdc_map_name```: ```rz_position_distortion_map```)?

jpienaar13:
Sorry now. Ill have something for you by Monday.

pdeperio:
@coderdj has cleaned up the file-based corrections in https://github.com/XENON1T/hax/pull/141.

So now I guess we need to think of how to implement the MC [here](https://github.com/XENON1T/hax/blob/e079a32107f52774d92f96c9a3fa472c698588e0/hax/treemakers/corrections.py#L75) and [here](https://github.com/XENON1T/hax/blob/e079a32107f52774d92f96c9a3fa472c698588e0/hax/treemakers/corrections.py#L58).

pdeperio:
Fixed by #144, but needs to be tested still.

PR #144, opened 2017-09-20
Updated Correction minitrees for MC compatibility

jpienaar13:
Changed corrections minitrees to look for a new tag in pax_metadata, 'mc_data', in order to track whether a pax file was produced from MC data or not. This is hopefully a resolution to #139.

If the file was generated from MC data, then it also attempts to pull the electron lifetime used from the pax_metadata

Hopefully this does its without breaking/affecting hax Corrections minitrees generated from real data, where the MC tag should be either non existent, or set to false if anyone does feel like adding it.

This, in addition with an update to MC/processing which I will commit in a bit, should make 
[XENON1T/pax#610](https://github.com/XENON1T/pax/pull/610) obsolete.


jpienaar13:
Add @pdeperio suggestions, the electron lifetime is now stored in hax.ini to allow for separate values to be specified during fax stage (where value is specified in SR*_parameters.ini), see [XENON1T/processing#43](https://github.com/XENON1T/processing/pull/43)

jpienaar13:
After some more input from @pdeperio, moved electron lifetime values into the correction_definitions dict in hax.ini (with some renaming, as looking for "file_names" no longer seemed entirely appropriate.)

Additionally, extended hax.runs.get_run_number(), to look in the pax_metadata for the MC tag, and if set to true, to use the MC run number. Corrections.py should now use this artificial MC run number when being made from MC pax files.

hax/hax.ini
```diff
@@ -111,23 +111,34 @@ sc_api_url = 'https://xenon1t-daq.lngs.infn.it/slowcontrol/GetSCData'
 sc_api_username = 'slowcontrolwebserver'
 
 ##
+# MC Electron Lifetime 
+##
+#SR dependent electron lifetime defined here for minitree production.
+#NB!: Electron lifetime is also specified in simulation initialization file. These might not be the same value.
+mc_electron_lifetime_liquid = {"sr0": 450, "sr1": 550}
```
pdeperio: Is this still needed?

hax/treemakers/corrections.py
```diff
@@ -53,6 +53,14 @@ class Corrections(TreeMaker):
 
     extra_metadata = hax.config['corrections_definitions']
 
+    #Check if data was generated with MC.
+    #If it was, pull the electron lifetime values from the associated metadata
+    def check_for_mc(self):
+        mc_data = False
+        if 'MC' in hax.paxroot.get_metadata(self.run_name)['configuration']:
```
pdeperio: Can this function be moved to ```runs.py```? (Looks like maybe some duplicated code there.)

hax/treemakers/corrections.py
```diff
@@ -92,7 +100,7 @@ def extract_data(self, event):
         # If there are no interactions cannot do anything
         if not len(event.interactions):
             return result
-
+
```
pdeperio: Watch out for flake8 whitespace

hax/treemakers/corrections.py
```diff
@@ -53,6 +53,14 @@ class Corrections(TreeMaker):
...
+    #If it was, pull the electron lifetime values from the associated metadata
+    def check_for_mc(self):
+        mc_data = False
+        if 'MC' in hax.paxroot.get_metadata(self.run_name)['configuration']:
+            mc_data = hax.paxroot.get_metadata(self.run_name)['configuration']['MC']['mc_generated_data']
+        return mc_data
+
     # Electron Lifetime: hopefully doc was pulled in hax.init.
     # Otherwise get it here at significantly higher DB cost
     try:
```
pdeperio: Should this check be skipped for MC?

hax/treemakers/corrections.py
```diff
@@ -144,7 +152,14 @@ def extract_data(self, event):
...
 
         # include electron lifetime correction
-        if self.elife_interpolation is not None:
+        pax_mc_data = self.check_for_mc()
+        if pax_mc_data:
+            wanted_electron_lifetime = self.get_correction("mc_electron_lifetime_liquid")
+            result['s2_lifetime_correction'] = np.exp((interaction.drift_time/1e3) /
+                                                      wanted_electron_lifetime)
+            print("This run is tagged as being MC data. Using MC electron lifetime value of %i us."
+                %wanted_electron_lifetime)
```
pdeperio: Add newline (flake8 issue) after here

hax/hax.ini
jpienaar13: Woops. No. Forgot to take that out.

hax/treemakers/corrections.py
jpienaar13: I didn't want it to throw an error if it checks for MC data in the configuration settings and doesn't find it there

jpienaar13:
I did check that I could produce corrections minitrees on real data yes.

hax/treemakers/corrections.py
jpienaar13: Maybe. The function in runs.py, checks if it is MC data as well, but is only used to set a run number, whereas this function is used to make the decision between using the interpolated lifetime or the correction init file electron lifetime

ErikHogenbirk:
FYI I am using this branch for some simulated AmBe data. In my simulated data, the 'MC' was not in the metadata, I suppose this will be fixed in a newer version of pax (or at the hax stage?)? I changed my local version of the code to force `mc_data = True` and then  it works like a charm. One comment, can you change the print statement (`"This run is tagged as being MC data. Using MC electron lifetime value of %i us."`) to a `log.debug`? It's printing messages for every dataset, and there are many of them for AmBe or NG MC.

coderdj:
This looks good. But don't you also want to be able to define custom files to use for the other corrections too? Or is this not necessary?

jpienaar13:
@ErikHogenbirk, once [XENON1T/processing#43](https://github.com/XENON1T/processing/pull/43) is merged, all future MC production should have the MC flag.

@coderdj I didn't follow your question. Which custom files where?

pdeperio:
@coderdj i guess you mean changing the map file names in ```hax.ini```?  then i think we can pass custom ```hax.ini```s through [hax.init("custom_hax.ini")](https://github.com/XENON1T/hax/blob/967003e54d4a3acd0e02e46b9241d8ab9dfe27d9/hax/__init__.py#L29) if necessary.

Issue #146, opened 2017-09-27
invalid value encountered in sqrt with new 3D FDC


pdeperio:
Following error observed in e.g. run 11199:
```
RuntimeWarning: invalid value encountered in sqrt
  result['z_new'] = -np.sqrt(result['z_observed_new']**2 - result['r_correction_new']**2)
```
@jingqiangye are you able to reproduce and fix?

Also, not sure if related, but job exited with
```
exceeded memory limit (1954804 > 1843200), being killed
```
(although Corrections minitree seems to have survived).

PR #147, opened 2017-09-30
Correct a small bug in 3D FDC

jingqiangye:
When making correction minitree in version 1.4, it'll show an error:
/project/lgrandi/anaconda3/envs/pax_head/lib/python3.4/site-packages/hax-1.7.0-py3.4.egg/hax/treemakers/corrections.py:238: RuntimeWarning: invalid value encountered in sort.

L237 I wanted to make sure absolute value of z_observed_new is larger than absolute value of r_correction_new, however in the current code I used r_correction_now directly as I forgot some of the value could be negative. Very stupid mistake... It's now corrected.

The bug now only affects few events in the very top.


pdeperio:
fixes #146 ?

jingqiangye:
Yes it'll fix #146

PR #148, opened 2017-10-02
Read pax metadata for MC only once during init

pdeperio:
To fix memory leak observed by cax daemon and reported on Gitter, introduced in #144.

Idea is to check MC flag from pax metadata only once, instead of every event, which seemed to be causing a memory leak (```get_metadata``` was returning a dict that wasn't being cleaned up?). Or @jpienaar13, did you have a reason to do this event-by-event?  If not, please check this branch and ensure it still works with your MC files.

(I will manually install this branch in ```pax_head```, ```pax_v6.8.0```, and ```pax_v6.8.1``` environments on Midway for now, so analyses and real data production (RDP) may continue.)


pdeperio:
merged so that master doesn't fall behind.  

@jpienaar13 still needs to check if this works with MC

jpienaar13:
Hi. I believe this is failing for MC data. Will push a fix ASAP.

PR #150, opened 2017-10-09
use uncorrected position for pax

feigaodm:
This PR is motivated by a change in [PAX, PR #620](https://github.com/XENON1T/pax/pull/620). With that change, we don't need to re-correct our x,y from a paried interactions.


pdeperio:
Is this superseded by https://github.com/XENON1T/hax/pull/151?

feigaodm:
yes, I think we can close this.

PR #151, opened 2017-10-10
Corrections Treemaker Update

pelssers:
I've been going through the Corrections Treemaker code to check which corrections are applied to positions.
There were some places where corrected positions were used which were then "decorrected" by applying a reverse transformation. Here its better to look up the uncorrected position from the ReconstructedPosition object right away instead of using the Interactions object which holds corrected positions.

I rewrote the Treemaker to do the following:

- Lookup the uncorrected positions from both TPF and NN (in the way described above).
- Add the old 2D FDC to the TPF positions (this is necessary since we have 2 different 2d FDC corrections as documented in hax.ini, if it were only 1 we could simply take the Interaction positions here.)
- Add the new 3D FDC to both the NN and TPF positions
- Added the loading of the new FDC map to the proper place (hax.ini)
- Cleaned up the loading of the different maps (still a bit messy but better).

In the docstring of the class you'll find the new naming convention for the various x,y variables. I tried to avoid using 'old/new' and instead went for '2d/3d', also explicitly adding which PosRec algorithms positions the correction is applied to.

I didn't find any test environment or unittests for hax so I have not been able to test this code, if you'll please let me know how changes to treemakers are tested I'll try and catch any bugs.


weiyuehuan:
Thanks Bart,  the name is much better now!  The x_obsreved_tpf which without the "decorrected" will be the true observed position from TPF. We need to check if it still have artificial feature with this observed position, if it shown better performance, we need to make new 3D FDC map, because NN and TPF observed position has [centimetres difference](https://xe1t-wiki.lngs.infn.it/lib/exe/detail.php?id=xenon%3Axenon1t%3Aanalysis%3Asr1%3Averifypax666&media=xenon:xenon1t:yuehuan:20170809:pax666_distance.png)

PR #152, opened 2017-10-13
Bug fix to allow self.mc check to succedd

jpienaar13:
Moved the location of self.mc check in hax/minitrees.py, as previously this was not updating the value to that read from the pax_metadata


feigaodm:
@jpienaar13 Have you test it with a simulation chain?

jpienaar13:
@feigaodm 

I have indeed tested it on MC data, and it solved the problem I was experiencing previously.

PR #154, opened 2017-10-17
Start using the new SR1 S2 correction map (v2.2)

adambrown1:
This causes the corrections treemaker to start using the v2.2 s2 map which is added to pax in pull request xenon1t/pax#626. The map is used for science run 1 data only (run number greater than 6386). For science run 0 the map is unchanged.

It should not be merged until the new map has been added to pax.


adambrown1:
@pdeperio done

Issue #155, opened 2017-10-18
Need multiple cS1 variables corresponding to each position?


pdeperio:
@berget2 derived new S1 LCE map using NN 3D FDC positions (https://github.com/XENON1T/pax/pull/628).  hax currently [creates only one ```cs1```](https://github.com/XENON1T/hax/blob/master/hax/treemakers/corrections.py#L254) variable from [the old map](https://github.com/XENON1T/hax/blob/master/hax/hax.ini#L134).  Should we make another ```cs1``` variable to correspond to the new map?

berget2:
Correction maps were added to run-db collections to unify/simplify applying different maps to different runs. With the FDC-3D using NN positions we're starting to build multiple sets of corrections for each set of runs. We should maintain a single set of corrections, or implement something else on the collections level to keep this tidy.

feigaodm:
If S1 map is defined as function of nn positions, in principle we should check whether it's applicapable to tpf positions. Not sure if we should create two cS1s at the stage from different algorithms for now. Maybe @pelssers can check whether we see some strange effect on the TPF position distributions using NN based FDC. 

I'm fine with keeping NN as the default maps for signal corrections, as I don't see position resolution is that significant in terms of S1&S2 variations. Which algrithms to use for FV definition could be another independent topic, if we can not prove NN FDC can be applied to TPF, probably we need to use NN as position indicator and TPF as a cut to ensure quality. But my understanding could be wrong.

pdeperio:
@berget2 I think when this Corrections treemaker was implemented, the RunsDB collection implementation became obsolete, right @coderdj?  i.e. Run dependence now handled in [hax.ini](https://github.com/XENON1T/hax/blob/master/hax/hax.ini#L134).

@feigaodm I agree.  How about https://github.com/XENON1T/hax/pull/156?

coderdj:
@pdeperio yeah we got rid of this run DB logic for everything except the electron lifetime correction. It's all in hax now and should be easier to oversee. That whole mess with all these DB entries was not a really well thought out system but we needed something to avoid bumping pax versions 10 times during the SR0 workshop.

As far as the subject of this thread. We should settle on a position then just have one correction. It's relatively easy to unapply and reapply custom corrections (with the other position for example) for people studying that kind of thing but I wouldn't necessarily formalize it.

pdeperio:
Implemented in #156

PR #156, opened 2017-10-20
New S1 LCE map based on NN 3D FDC

pdeperio:
Update correction for S1 based on https://github.com/XENON1T/pax/pull/628. 

This switches the default ```cs1``` to use NN 3D FDC as proposed in https://github.com/XENON1T/hax/issues/155.  The old TPF 2D FDC ```cs1``` moved to ```cs1_tpf_2dfdc ``` for comparison studies.

Awaiting SR0 map before merging.

(Warning: not tested yet.)


pdeperio:
@tunnell @feigaodm Why is codacy disabled now?

pdeperio:
then lax should be updated explicitly

feigaodm:
yeah, a change in lax will be good enough. Ideally we should use the 'best' algrithm as the default 'x,y,z' positions and right now it's NN after 3d-FDC. But this can be updated whenever a better one is chozen.

pdeperio:
SR0 LCE map now implemented from https://github.com/XENON1T/pax/pull/631

OK from my side, unless @feigaodm or @coderdj want to make position variable names more explicit or remove ```cs1_tpf_2dfdc``` (i.e. make final decision on ```cs1``` using NN 3D FDC now).

PR #157, opened 2017-10-24
Get extra info on L.O. S2 for Kr83m removal

zhut19:
Initial commit on getting pattern, time delay info for Kr83m miss classified S1 removal


feigaodm:
It shouldn't matter, but standard dt is calculated by peak.index_of_maximum instead of hit_time_mean

PR #158, opened 2017-10-25
Add time-dependent 3D FDC for SR1

jingqiangye:
Update the map for time-dependent 3D FDC.


feigaodm:
More info can be found from [PAX PR632](https://github.com/XENON1T/pax/pull/632)

PR #161, opened 2017-11-07
update blinding cut to new 2-sigma ER line

coderdj:
2-sigma ER taken from Shingo's note here: https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:kazama:anomalous_bg_sr1#appendix_2er_nr-band_lines_estimated_with_pax_v680_3d_fdc_and_1300_kg_fv

Note that this changes the definition to cs2_bottom instead of cs2 as it was.

I tested this quickly on a couple runs of rn220_tail data with no cuts, which should be safe, and it looks good. It would be nice if a second person tested as well before deploying.

### DIFF
--- a/hax/hax.ini
+++ b/hax/hax.ini
@@ -99,7 +99,7 @@ trigger_data_special_path = None
 ##
 # Blinding cut
 ##
-blinding_cut  = '(log(cs2_bottom/cs1)/log(10) > 0.374125*exp(-cs1/33.5861) + 1.43559 -0.000755488*cs1 + 1.33334/cs1 | (cs1 > 200) | (s2<150) | (largest_other_s2>200)'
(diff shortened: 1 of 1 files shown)


coderdj:
Piggy-backing onto previous commit to add new blinding for ER analyses. 

The region from 50-80 keV is blinded for the 2e- capture analysis and the region from 2300-2600 keV is blinded for the 2vbb analysis. Especially in the latter case this is certainly too large, but since I'm not sure of our energy resolution in that region it's safe for now and the analysts responsible can suggest a more suitable range.

Here's a simple plot showing the spectrum for some background runs with this blinding cut, r_3d_nn<42, and cs1>200 applied. Holes seem to be in the right spot. 

ces_blind

feigaodm:
nice change! I agree to merge it for now. Later next week we probably will have another update regarding what should be blinded even in higher energy region. Say we will only blind events between NR-3sigma and ER-2sigma. And in the ER band, we will blind everything accoring to the Science searches we want, like the ones added here for double electron capture and 0vbb searches.

coderdj:
Wait! I screwed up the CES cut. There was an errant "/100" in there. The new plot looks correct. At least we see peaks.

cesbetter

PR #162, opened 2017-11-07
Don't query db for every run's blinding status

JelleAalbers:
When you load all SR0 bg datasets (~4000), you have to wait ~9 minutes before hax starts loading the minitrees. During this time it's querying the runs db (one query / run) to fetch some info for the decision whether or not to apply the blinding cut (name, number, tags). However, hax has already queried this info when you do hax.init(). This fixes hax to use that info instead, cutting the initialization time of the minitrees load call down to a few seconds at most.

I also changed get_run_number to raise an exception rather than return 0 when it cannot find a run.

Test code (100 datasets):

    %%time
    results = np.zeros(100)
    for i, run_i in enumerate(6666 + np.arange(100)):
        results[i] = hax.runs.is_blind(run_i)
    print("%d are blind" % results.sum())

Before:

    81 are blind
    CPU times: user 588 ms, sys: 38.1 ms, total: 626 ms
    Wall time: 14.3 s

After:

    81 are blind
    CPU times: user 312 ms, sys: 12.5 ms, total: 325 ms
    Wall time: 318 ms

### DIFF
--- a/hax/cuts.py
+++ b/hax/cuts.py
@@ -203,7 +203,7 @@ def range_cuts(*args, **kwargs):
     range_selections(*args, **kwargs)
 
 
-def apply_lichen(data, lichen_names, lichen_file='sciencerun0', **kwargs):
+def apply_lichen(data, lichen_names, lichen_file='sciencerun1', **kwargs):
     """Apply cuts defined by the lax lichen(s) lichen_names from the lichen_file to data.
     """
     # Support for single lichen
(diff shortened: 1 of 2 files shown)


JelleAalbers:
Also pushed an update to hax.cuts.apply_lichen to default to sciencerun1 and not complain when cuts have strings as versions (not sure if people use this function though).

PR #163, opened 2017-11-10
Corrected Signals for the DoubleScatter Minitree

ctherreau:
New DoubleScatter Treemaker with Corrected NN positions and Corrected S1 signals 

Check of this new Treemaker [here](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenon1t:chloe:correcteddoubles1scatter_firstcheck)


hax/treemakers/CorrectedDoubleS1Scatter.py
```diff
@@ -0,0 +1,386 @@
...
+        
+    #### Correction Map
+    
+        # Check that the correct S2 map is loaded and change if not
+        self.xy_map, self.loaded_xy_map_name = self.load_map("s2_xy_map",
+                                                             self.xy_map,
+                                                             self.loaded_xy_map_name)
+
+        # Load the 2D FDC map
+        self.fdc_2d_map, self.loaded_2d_fdc_map_name = self.load_map("fdc_2d",
```
coderdj: Also not using this I think

hax/treemakers/CorrectedDoubleS1Scatter.py
```diff
@@ -0,0 +1,386 @@
...
+        result['int_b_y_pax']=interactions[int_b].y
+        result['int_b_z_pax']=interactions[int_b].z
+
+######### Compute DoubleScatter Specific Variables #########
+            
+        # Select largest hits on each channel in s10 and s11 peaks
+        s1_a_hitChannels = []
+        s1_a_hitAreas = []
+        
+        for hit in peaks[s1_a].hits:
```
coderdj: Isn't this a bit verbose? The thing you want in the end is the number of distinct channels. There are ways to get there quicker i.e. https://docs.scipy.org/doc/numpy-1.13.0/reference/generated/numpy.setdiff1d.html

hax/treemakers/CorrectedDoubleS1Scatter.py
```diff
@@ -0,0 +1,386 @@
...
+                result['int_a_r_nn'] = np.sqrt(rp.x ** 2 + rp.y ** 2)
+                
+        for rp in peaks[s2_b].reconstructed_positions:
+            if rp.algorithm == 'PosRecNeuralNet':
+                result['int_b_x_nn'] = rp.x
+                result['int_b_y_nn'] = rp.y
+                result['int_b_r_nn'] = np.sqrt(rp.x ** 2 + rp.y ** 2)
+
+
+        int_a_z = interactions[int_a].z - interactions[int_a].z_correction
```
coderdj: This is z_observed (uncorrected) right? Should it be marked int_a_z_observed?

hax/treemakers/CorrectedDoubleS1Scatter.py
```diff
@@ -0,0 +1,386 @@
...
+        
+        result['cs1_a_int_a'] = peaks[s1_a].area * result['s1_int_a_xyz_correction_nn_fdc_3d']
+        result['cs1_b_int_a'] = peaks[s1_b].area * result['s1_int_a_xyz_correction_nn_fdc_3d']
+
+        
+        result['s1_int_b_xyz_correction_nn_fdc_3d'] = 1 / self.lce_map_nn_fdc_3d.get_value(result['int_b_x_3d_nn'],
+                                                                                     result['int_b_y_3d_nn'],
+                                                                                     result['int_b_z_3d_nn'])
+        
+        result['cs1_b_int_b'] = peaks[s1_b].area * result['s1_int_b_xyz_correction_nn_fdc_3d']
```
coderdj: Missing cs1_a_int_b? 

I have to admit the naming confuses me but as long as the people using the tree can keep it straight that's fine.

hax/treemakers/CorrectedDoubleS1Scatter.py
ctherreau: Yes you're right. Did it

hax/treemakers/CorrectedDoubleS1Scatter.py
ctherreau: Yes, I change it by making the "int_a" correction the default one (cS1_a and cS1_b). 
Only "int_b" correction for S1_b signal is called with cs1_b_int_b. 
There is no cS1_a_int_b because it makes no sense to correct S1_a signal using int_b position.

hax/treemakers/CorrectedDoubleS1Scatter.py
```diff
@@ -0,0 +1,336 @@
...
+                      'interactions.z_correction',
+                      'interactions.drift_time',
+                      'start_time',
+                      'peaks.hits*',
+                      'interactions.s1_pattern_fit'
+                     ]
+    extra_metadata = hax.config['corrections_definitions']
+    
+    # Load Correction Map
+    loaded_3d_fdc_map_name = None
```
pdeperio: Should we make a common "maps" class that takes care of all of these at hax init? (Reducing the duplicated functions below also in ```corrections.py```.)

hax/treemakers/CorrectedDoubleS1Scatter.py
berget2: There's certainly more straightforward ways, this just required the least thought when I originally put this together. Maybe better to move forward with adding corrections for now and consider a more thorough re-write soon.

hax/treemakers/CorrectedDoubleS1Scatter.py
```diff
@@ -0,0 +1,336 @@
+# Get pax values
+import pax
+from pax import configuration
+pax_config = configuration.load_configuration('XENON1T')
+tpc_length = pax_config['DEFAULT']['tpc_length']
+tpc_radius = pax_config['DEFAULT']['tpc_radius']
```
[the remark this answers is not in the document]
berget2: agreed

hax/treemakers/CorrectedDoubleS1Scatter.py
ctherreau: I tried to but I am still quite new in Python and I didn't find a way to make it works

coderdj: @pdeperio yes, but considering the time sensitivity we thought to do it after. For example next time we need to make an update we can just request a refactor, hopefully at a point when we can afford to spend the extra week

PR #164, opened 2017-11-13
Create flash_identification.py

oliw26:
New Treemaker which identifies if there was a flash during a run. It contains additional information about the flash itself and the time distance of each event.


hax/treemakers/flash_identification.py
```diff
@@ -0,0 +1,94 @@
...
+
+        #Get BUSY_ON (channel id 255)
+        self.BUSY_data = self.transposed[255]
+
+        time_window = []
+        window_biggest = []
+
+        # Check in BUSY_ON channel if there where an increase "typical" for flashes and get the time-window in seconds
+        for t in range(0,len(self.BUSY_data)):
+            if self.BUSY_data[t] > 20:
```
coderdj: Why 20? Shouldn't a lower number also indicate a problem? Like 5?

Wondering if some smaller flashes won't hit this threshold

hax/treemakers/flash_identification.py
```diff
@@ -0,0 +1,94 @@
...
+            # check if there was also a "large" increase in one of the PMT channels
+            self.flash_time_highest_trig = int(np.argmax(trigger_data)/len(self.transposed))
+
+            # check if the two incidence match (roughly) in time
+            if abs(self.flash_time_highest_trig - self.flash_time_highest_trig) < 200:
+                self.flashing_PMT = np.argmax(trigger_data[self.flash_time_highest_trig])
+                self.flash_amplitude = int(np.max(self.transposed[self.flashing_PMT]))
+
+                # all flashes found so far a well beyond this threshold
+                if self.flash_amplitude < 20000:
```
coderdj: Why is this threshold needed? Do you find other things where the amplitude is, say, 10,000?

oliw26: Yes we have some spikes in the trigger_data were a channel exceeds 10,000. These PMTs are not known to have any problems. Sometimes afterwards also there is also a DAQ issue. i.e. run 11482 channel 24. Its not a flash so I rejected it. If we also want to get rid of these features one can think about lowering the threshold.
For now all flashes are well above 20,000. However now it seems that the flashes get smaller with time. However this threshold should still hold for a while.

hax/treemakers/flash_identification.py
oliw26: All flashes I found were above this threshold. It does also not change over time.
If I use a lower threshold I get quite some other issues, which would lead in quite some loss of time if I cut a bigger time-window around all of them. Is this not already accounted for in the DAQ cut?

coderdj: Should be. Was just curious what kinds of things you found. I only see a handful of runs in SR1 with DAQ issues. SR0 is much higher though so probably this threshold makes sense for extending to SR0 too.

coderdj:
I'm fine with changes. Interested to see how plots look after processing a bunch of runs.

PR #168, opened 2017-11-16
Update blinding_cut for high energy and NR band

pdeperio:
Blind the NR band outside currently blinded region of ```cs1<200``` for:

* higher energies, for e.g. [inelastic DM](https://arxiv.org/abs/1608.02662) searches (requested by @feigaodm), and
* preparation for [wall leakage sideband unblinding](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenon1t:analysis:meetings:20171113:sideband#unblinding_proposal) proposed by @jingqiangye, to avoid potential signal contamination in model building.

In addition to SR1 Rn220, AmBe, and NG data (selected by an extension of @skazama's [framework](https://github.com/XENON1T/SR1Results/tree/master/Calibration/inputs_for_bbf)), @mcfatelin provided a flat NR spectrum up to 350 keV simulated through the [bbf template generator](https://github.com/XENON1T/bbf/pull/37), assuming the [NEST model](https://arxiv.org/pdf/1412.4417.pdf) with data up to ~200 keV and extrapolated beyond as shown in the following figure.
nest_nr_cy

This simulation is shown in the following figure, together with AmBe+NG data points.

nr_signal_compare
A Gaussian is fitted to each ```cs1``` slice and the -4.5 sigma points are used to derive the parameterization. This larger than usual sigma is chosen to cover:
* the fact that the lower tails are larger than Gaussian,
* data mean starts to deviate down at high energies, and
* the imperfect fit.

Below ```cs1=20```, a constant is assumed to cover the spectral variation of low mass WIMPs, as shown in the following figure, together with the ER band parameterization assuming a constant above ```cs1=500``` (updating the previously defined curve by @skazama in https://github.com/XENON1T/hax/pull/161).

blinding_highe

Below the NR line for ```cs1>200``` is left unblinded as before, however above will be newly blinded and may affect the previous [wall leakage analysis](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenon1t:analysis:meetings:20171113:sideband), so should be checked by @jingqiangye (even before merging this by manually applying the cut line).

This new blinding cut is tested on some SR0 background data in the following figure and looks fine (lower band is below S2 threshold).
bkg_test_blinding

Together with sideband unblinding, we need to ensure that the NR band remains blinded down to low energies since the parameterization is currently implemented for ```cs1>200```, so should lower this threshold and then also add the constant part ```((cs1<20)&(log(cs2_bottom/cs1)/log(10) < 1.08159))```.


jingqiangye:
The region below NR band for cs1 between 200 and 500 PE is kept, and the s2 in this region is enough to cover s2 for wall leakage events in ROI, so it won't affect wall background model.

cwittweg:
This cut removes some events from the Kr-83m line which reaches out of the band due to its two-fold decay. This is the 2D fit with the blinding applied:
kr83m_fit_no_blinding
Here it is without.
kr83m_fit_with_nr_he_blinding
The result of the 2D fit (for determining g1 and g2) is only changed for this line. The values of g1 and g2 change: 
g1 = 0.1426 => g1 = 0.1427 (error 0.0001)
g2 = 11.36 => g2 = 11.38 (error 0.01)

Could we maybe move the blinding down a bit, so it does not truncate the Kr-83m ellipse? The other lines are fine. Energy spectrum does not change significantly.

pdeperio:
Thanks for the comments! I've updated the ER line by including Kr83m calibration data -3*RMS points as shown in the following figure, and correspondingly lowering the high energy constant:

blinding_highe

and compared to the NR signal simulation in this figure:

nr_signal_compare

and again tested on some SR0 background data:

bkg_test_blinding

pdeperio:
Just checking blinding lines again with SR0 calibration data:

blinding_highe_s2thresh_sr0

(Open this and previous image in new tabs to compare.) Confirming lines are close enough to use for SR0 too.

PR #170, opened 2017-11-21
Update CorrectedDoubleS1Scatter.py

ctherreau:
Update with S2_a corrections (x-y and e-lifetime corrections)

Some plots to check the corrections on the Update here : https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenon1t:chloe:correcteddoubles1scatter_firstcheck#Update-:-S2-corrections


ctherreau:
> Thanks Chloe, code looks fine. (Just another reminder that we should refactor these reused functions and objects later.)
For a next step to your note, could you please extend the fiducial volume (FV) and re-run? We would like to confirm that all our corrections our working properly as we attempt to expand our FV. For example, plots showing the cS1 and cS2 resolutions as a function of radial slices up to the TPC radius (47.9 cm). Also, we may be expanding in Z too as proposed by Sander here.
Now that you have cS2, can you also reconstruct the energy using g1 and g2?

@pdeperio 
Sure, I will add this to my note. 
Thanks !

PR #172, opened 2017-12-01
Update version for e-lifetime update

pdeperio:
Update version in ```Corrections``` and ```CorrectedDoubleS1Scatter``` treemakers to regenerate minitrees after @zachgreene updated e-lifetime function to v0.4 in DB, corresponding to model 3.10 [here](https://xe1t-wiki.lngs.infn.it/doku.php?id=greene:electron_lifetime_attempts) (SR0 should be unaffected, i.e. he implemented a piece-wise function stepping in between SRs).


zachgreene:
I recalculated the  [Kr-83m correction](https://xe1t-wiki.lngs.infn.it/doku.php?id=greene:electron_lifetime_attempts) so the evolution changed slightly.  Plots for the evolution over SR1 and since operations are shown.
piecewise_170912_171126cont_beginning
piecewise_170912_171126cont_sr1

pdeperio:
Updated on Midway ```pax_v6.8.0``` environment only so far.

PR #174, opened 2017-12-07
New reconstruction and dependent variables

coderdj:
This adds the s1_pattern_fit using 3d FDC and neural network posrec. The other version uses the 2D FDC with TPF. This also adds the s1 area fraction top probability variable computed at the 3D-NN position.

Diagnostic plots incoming. Putting this here as placeholder.


hax/treemakers/posrec.py
```diff
@@ -0,0 +1,163 @@
...
+                      'peaks.hits_per_channel[260]',
+                      'peaks.n_saturated_per_channel',
+                      'interactions.x','interactions.y','interactions.z']
+
+    def __init__(self):
+        
+        hax.minitrees.TreeMaker.__init__(self)
+        
+        # We need to pull some stuff from the pax config
+        self.pax_config = load_configuration("XENON1T")
```
pdeperio: do we also need SR-dependent gains from configs in processing repo?

hax/treemakers/posrec.py
```diff
@@ -0,0 +1,163 @@
...
+        # load trained NN models
+        nn_model_json = utils.data_file_name('tensorflow_nn_pos_XENON1T_20171211.json')
+        json_file_nn = open(nn_model_json, 'r')
+        loaded_model_json = json_file_nn.read()
+        json_file_nn.close()
+        loaded_nn_model = model_from_json(loaded_model_json)
+        weights_file = utils.data_file_name('tensorflow_nn_pos_weights_XENON1T_20171211.h5')
+        loaded_nn_model.load_weights(weights_file)
+        self.nn_tensorflow = loaded_nn_model
+        self.list_bad_pmts = [1, 2, 12, 26, 34, 62, 65, 79, 86, 88, 102, 118, 130, 134, 135, 139, 148, 150, 152, 162, 178,
```
pdeperio: can this be taken from the PMT gain list above?

hax/treemakers/posrec.py
```diff
@@ -0,0 +1,163 @@
...
+                         183, 190, 198, 206, 213, 214, 234, 239, 244]
+        self.ntop_pmts = 127 # How to get this automatically?
```
pdeperio: indeed must be automated for SR dependency (following above comments)

hax/treemakers/posrec.py
```diff
@@ -0,0 +1,163 @@
...
+            adjust_to_qe=qes[self.tpc_channels],
+            default_errors=(self.pax_config['DEFAULT']['relative_qe_error'] +
+                            self.pax_config['DEFAULT']['relative_gain_error'])
+        )
+
+        # For s1_aft variable
+        # Don't yell at me for hardcoding the filename into hax because it was
+        # also hardcoded in pax. Just kicking the can.
+        #aftmap_filename = utils.data_file_name('s1_aft_xyz_XENON1T_06Mar2017.json')
+        aftmap_filename = utils.data_file_name('s1_aft_xyz_XENON1T_20170808.json')
```
pdeperio: move to ```hax.ini``` like other maps in ```corrections.py```?

hax/treemakers/posrec.py
```diff
@@ -0,0 +1,163 @@
...
+        self.aft_map = InterpolatingMap(aftmap_filename)
+        self.low_pe_threshold=10
+        #hax.minitrees.Treemaker.__init__(self)
+        # load trained NN models
+        nn_model_json = utils.data_file_name('tensorflow_nn_pos_XENON1T_20171211.json')
```
pdeperio: move to ```hax.ini``` like other maps in ```corrections.py```

hax/treemakers/posrec.py
```diff
@@ -0,0 +1,163 @@
...
+        s1 = event.peaks[interaction.s1]
+        s2 = event.peaks[interaction.s2]
+
+        # Position reconstruction based on NN from TensorFlow
+        s2apc = np.array(list(s2.area_per_channel))
+        s2apc_clean = []
+        for i, s2_t in enumerate(s2apc):
+            if i not in self.list_bad_pmts and i < self.ntop_pmts:
+                s2apc.clean.append(s2_t)
+        s2acp_clean.np.assaray(s2acp_clean)
```
pdeperio: acp -> apc?

hax/treemakers/posrec.py
coderdj: If I understand your question correctly then no. Gains already applied at this stage. This is just to pull some defaults. Unfortunately we wouldn't get override values in case the defaults are overridden in the run doc. I checked that they are not overridden for the values I took but I'm not sure of a better way to generalize.

hax/treemakers/posrec.py
coderdj: Why needed? Bad pmts are blinded in pax (gain zero) so have no signal in the hit pattern.

hax/treemakers/posrec.py
coderdj: Same comment as above. You should just be able to get this from the XENON1T config because XENON1T will always have the same number of PMTs in the top array. Sure, some will be blinded, but they always have zero signal then.

hax/treemakers/posrec.py
```diff
@@ -131,7 +130,9 @@ def extract_data(self, event):
         aft = self.aft_map.get_value(self.x[i], self.y[i], self.z[i])
         event_data['s1_area_fraction_top_probability_hax'] = binom_test(
             size_top, size_tot, aft)
-
+        event_data['s1_area_fraction_top_probability_hax2'] = binom_test_pax(
```
pdeperio: @darrylmasson @coderdj what's the difference with this one? (better name than ```hax2```?)

darrylmasson: The binom_test from scipy is only valid for integer inputs of PE. The version I put into pax accepts floating points and is more accurate. I don't think there's a particularly good reason to include the scipy-based calculation (unless we want to compare with the original SR0 codes)

hax/treemakers/posrec.py
```diff
@@ -0,0 +1,173 @@
...
+
+        try:
+            event_data['s1_pattern_fit_hax'] = self.pattern_fitter.compute_gof(
+                (self.x[i], self.y[i], self.z[i]),
+                apc[self.tpc_channels],
+                pmt_selection=np.setdiff1d(self.tpc_channels,
+                                           confused_s1_channels),
+                statistic=self.statistic)
+
+            event_data['s1_pattern_fit_hits_hax'] = self.pattern_fitter.compute_gof(
```
pdeperio: this (hits) is using the same MC map as for area?

hax/treemakers/posrec.py
```diff
@@ -0,0 +1,173 @@
...
+                statistic=self.statistic)
+
+            pmts_bottom = np.setdiff1d(self.tpc_channels, confused_s1_channels)
+            pmts_bottom[0:127] = 0
+            event_data['s1_pattern_fit_bottom_hax'] = self.pattern_fitter.compute_gof(
+                (self.x[i], self.y[i], self.z[i]),
+                hpc[self.tpc_channels],
+                pmt_selection=pmts_bottom,
+                statistic=self.statistic)
+
```
pdeperio: add also variable for hits as above?

hax/treemakers/posrec.py
coderdj: Yuck. OK I had them both for testing. The scipy one fails miserably below ~10pe. The plan is to revert to one before merging. The one will be the pax one I guess.

hax/treemakers/posrec.py
coderdj: In pax yes but we don't actually use hits so I don't know why I included it.

pdeperio: might be ok to keep, since the optical maps are using photons detected i think, not PE

pdeperio:
@pelssers have you seen this yet? wondering if the treemaker you were working on is in a more advanced state (like how you made ```corrections.py``` more general for all reconstruction algorithms).

hax/treemakers/posrec.py
pdeperio: fixed in https://github.com/XENON1T/hax/pull/180

hax/treemakers/posrec.py
```diff
@@ -0,0 +1,173 @@
...
+            return event_data
+        interaction = event.interactions[0]
+        s1 = event.peaks[interaction.s1]
+        s2 = event.peaks[interaction.s2]
+
+        # Position reconstruction based on NN from TensorFlow
+        s2apc = np.array(list(s2.area_per_channel))
+        if(len(s2apc)!=self.ntop_pmts):
+            return event_data
+        s2apc_clean = []
```
pdeperio: @feigaodm @weiyuehuan: as @coderdj mentioned above, is this needed if ```s2apc``` is already 0 from pax for bad PMTs? or does ```nn_tensorflow.predict``` actually need to remove these channels from the list?

hax/treemakers/posrec.py
```diff
@@ -0,0 +1,173 @@
...
+            i = self.indices.index(e)
+        except Exception:
+            return event_data
+        interaction = event.interactions[0]
+        s1 = event.peaks[interaction.s1]
+        s2 = event.peaks[interaction.s2]
+
+        # Position reconstruction based on NN from TensorFlow
+        s2apc = np.array(list(s2.area_per_channel))
+        if(len(s2apc)!=self.ntop_pmts):
```
pdeperio: @feigaodm @weiyuehuan: in what case does this actually happen? i.e. is it ok to just ignore these events?

coderdj:
I'm starting some cleanup of this but it seems the NN stuff doesn't work at all. Did it ever work?

First it won't even import on midway login nodes due to some glibc error, but we can live with this for now since you can run it on batch nodes fine. Then line https://github.com/XENON1T/hax/blob/newpatternlikelihood/hax/treemakers/posrec.py#L114 causes, by definition, a return before the neural net is ever invoked. After removing that clause I start getting weird tensorflow-related errors.  

Can someone say if this is working for them in some setup? I don't mind helping debugging but we should be sure to test before committing.

pdeperio:
Worked for me before. But @feigaodm and I were thinking that posrec stuff should be moved to another treemaker, agree?

coderdj:
Oh I just got it working. We were clearing the keras session in the init, which wiped the network we just loaded. I'll commit the fix.
Your mass production all failed unfortunately.

coderdj:
As for splitting into another minitree. May be useful for testing, but unpacking the peak objects and hit patterns is computationally slow so it might make sense to have everything that needs to do that in a single minitree.

skazama:
The list of excluded PMTs is different between SR0 and SR1. 
For SR0, it is [1, 12, 26, 27, 34, 51, 62, 65, 73, 79, 86, 88, 91, 102, 118, 130, 134, 135, 137, 139, 148, 150, 152, 162, 167, 178, 183, 198, 203, 206, 213, 214, 234, 239, 244].

hax/treemakers/posrec.py
```diff
@@ -0,0 +1,200 @@
...
+                      'peaks.n_saturated_per_channel[260]',
+                      'interactions.x', 'interactions.y', 'interactions.z']
+
+    def __init__(self):
+
+        hax.minitrees.TreeMaker.__init__(self)
+        
+        # We need to pull some stuff from the pax config
+        self.pax_config = load_configuration("XENON1T")
+        self.tpc_channels = list(range(0, 247 + 1))
```
tunnell: You can fetch this from pax config in previous line

hax/treemakers/posrec.py
```diff
@@ -0,0 +1,200 @@
...
+        )
+
+        # For s1_aft variable
+        aftmap_filename = utils.data_file_name(hax.config["s1_aft_filename"])
+        self.aft_map = InterpolatingMap(aftmap_filename)
+        self.low_pe_threshold = 10
+
+        # load trained NN models
+        nn_model_json = utils.data_file_name(hax.config["neural_network_model"])
+        json_file_nn = open(nn_model_json, 'r')
```
tunnell: This is the bad way to load a model.  Please do `model.save('blah.hdf5')`, which will save the model and weights together. There's no reason to keep these apart and it can just lead to problems.

tunnell: Ah, there's also a keras.utils load_model command to reload it.

hax/treemakers/posrec.py
```diff
@@ -0,0 +1,200 @@
...
+        loaded_model_json = json_file_nn.read()
+        json_file_nn.close()
+        loaded_nn_model = model_from_json(loaded_model_json)
+        weights_file = utils.data_file_name(hax.config['neural_network_weights'])
+        loaded_nn_model.load_weights(weights_file)
+        loaded_nn_model.compile(loss='mean_squared_error', optimizer='adam')
```
tunnell: If you load the model as mentioned before, you don't need to recompile.  You shouldn't need to recompile anyways since you're not retraining it, right?

hax/treemakers/posrec.py
```diff
@@ -0,0 +1,200 @@
...
+        self.nn_tensorflow = loaded_nn_model
+        self.list_bad_pmts = [1, 2, 12, 26, 34, 62, 65, 79, 86, 88, 102, 118,
+                              130, 134, 135, 139, 148, 150, 152, 162, 178, 183,
+                              190, 198, 206, 213, 214, 234, 239, 244, 27, 73,
+                              91, 137, 167, 203]
+        self.ntop_pmts = 127  # How to get this automatically?
```
tunnell: https://github.com/XENON1T/pax/blob/master/pax/config/XENON1T.ini#L450
`
len(pax_config['channels_top'])
`

hax/treemakers/posrec.py
```diff
@@ -0,0 +1,200 @@
...
+                
+    def get_data(self, dataset, event_list=None):
+
```
tunnell: Docstring?  What is this used for?

hax/treemakers/posrec.py
```diff
@@ -0,0 +1,200 @@
...
+            "s1_pattern_fit_hax": None,
+            "s1_pattern_fit_hits_hax": None,
+            "s1_pattern_fit_bottom_hax": None,
+            "s1_pattern_fit_bottom_hits_hax": None,
+            "s1_area_fraction_top_probability_hax": None,
+            "x_observed_nn_tf": None,
+            "y_observed_nn_tf": None,
+            "s1_area_upper_injection_fraction": None,
+            "s1_area_lower_injection_fraction": None,
+        }
```
tunnell: I sort of feel that each category of variable should be its own subfunction.  This is super hard to follow.

hax/treemakers/posrec.py
```diff
@@ -0,0 +1,200 @@
...
+
+        # Now do s1_pattern_fit
+        apc = np.array(list(s1.area_per_channel))
+        hpc = np.array(list(s1.hits_per_channel))
+
+        # Get saturated channels
+        confused_s1_channels = []
+        for a, c in enumerate(s1.n_saturated_per_channel):
+            if c > 0:
+                confused_s1_channels.append(a)
```
tunnell: Can be one line https://github.com/XENON1T/pax/blob/master/pax/datastructure.py#L249

hax/treemakers/posrec.py
```diff
@@ -0,0 +1,200 @@
...
+
+            event_data['s1_pattern_fit_hits_hax'] = self.pattern_fitter.compute_gof(
+                (self.x[i], self.y[i], self.z[i]),
+                hpc[self.tpc_channels],
+                pmt_selection=np.setdiff1d(self.tpc_channels,
+                                           confused_s1_channels),
+                statistic=self.statistic)
+
+            pmts_bottom = np.setdiff1d(self.tpc_channels, confused_s1_channels)
+            pmts_bottom[0:127] = 0
```
tunnell: Hardcode, though you define before 127.

tunnell: Comment on how to get this list of PMTs above

hax/treemakers/posrec.py
```diff
@@ -0,0 +1,200 @@
...
+            size_tot = s1.n_hits * (1. - s1_frac) + s1.area * s1_frac
+        else:
+            size_top = s1.area * s1.area_fraction_top
+            size_tot = s1.area
+
+        aft = self.aft_map.get_value(self.x[i], self.y[i], self.z[i])
+        event_data['s1_area_fraction_top_probability_hax'] = binom_test_pax(
+            size_top, size_tot, aft)
+
+        # Now do s1_pattern_fit
```
tunnell: apc? hpc?

hax/treemakers/posrec.py
```diff
@@ -0,0 +1,200 @@
...
+        s2apc_clean = []
+        for i, s2_t in enumerate(s2apc):
+            if i not in self.list_bad_pmts and i < self.ntop_pmts:
+                s2apc_clean.append(s2_t)
+        s2apc_clean = np.asarray(s2apc_clean)
+        s2apc_clean_norm = s2apc_clean / s2apc_clean.sum()
+        s2apc_clean_norm = s2apc_clean_norm.reshape(1, len(s2apc_clean_norm))
+        predicted_xy_tensorflow = self.nn_tensorflow.predict(s2apc_clean_norm)
+        event_data['x_observed_nn_tf'] = predicted_xy_tensorflow[0, 0] / 10.
+        event_data['y_observed_nn_tf'] = predicted_xy_tensorflow[0, 1] / 10.
```
tunnell: Can this preprocessing be broken into its own function?  If I am to add another model, then it's copy-paste 10 more lines... though the preprocessing is slightly different.

Also, clear is the wrong word.  It's 'preprocessed' or 'normalized'.

feigaodm: @tunnell It looks like you have some nice ideas on how to clean up the codes here. Can you modify this PR and test it on a dataset. Maybe you can also fix the issue that sometimes TF fails to be running.

hax/treemakers/posrec.py
pdeperio: Fixed in https://github.com/XENON1T/hax/pull/174/commits/d61afe00aa7246043a4fd1b58e7bbf1cef93ce48

hax/treemakers/posrec.py
pdeperio: Fixed in https://github.com/XENON1T/hax/pull/174/commits/d61afe00aa7246043a4fd1b58e7bbf1cef93ce48

hax/treemakers/posrec.py
pdeperio: Fixed in https://github.com/XENON1T/hax/pull/174/commits/d61afe00aa7246043a4fd1b58e7bbf1cef93ce48

hax/treemakers/posrec.py
pdeperio: Fixed in https://github.com/XENON1T/hax/pull/174/commits/ef6fd7de164d451d2e3ee127b35765c72246b28b

hax/treemakers/posrec.py
pdeperio: https://github.com/XENON1T/hax/pull/174/commits/ef6fd7de164d451d2e3ee127b35765c72246b28b

hax/treemakers/posrec.py
pdeperio: Fixed in https://github.com/XENON1T/hax/pull/174/commits/d61afe00aa7246043a4fd1b58e7bbf1cef93ce48

hax/treemakers/posrec.py
pdeperio: Fixed in https://github.com/XENON1T/hax/pull/174/commits/c8d44aa448b7571906c6e5d2a2e645be2d870640

hax/treemakers/posrec.py
```diff
@@ -0,0 +1,269 @@
...
+                if 'correction' in entry:
+                    return entry['correction']
+
+                # For NN files
+                elif 'weights' in entry and 'model' in entry:
+                    return entry['weights'], entry['model']
+
+        return None
+
+    def load_map(self, name, loaded_map, loaded_name):
```
coderdj: Seriously? Sorry this is now getting way too sloppy. This code is straight up copy-pasted out of a different treemaker.

hax/treemakers/posrec.py
pdeperio: @feigaodm says: defines the number of layers in NN.

Hardcoded for SR0 and SR1 now in https://github.com/XENON1T/hax/pull/184. @weiyuehuan working on including this list with the NN models to replace this.

hax/treemakers/posrec.py
pdeperio: Fixed in https://github.com/XENON1T/hax/pull/180

hax/treemakers/posrec.py
pdeperio: See response above and https://github.com/XENON1T/hax/pull/184

hax/treemakers/posrec.py
pdeperio: Fixed in https://github.com/XENON1T/hax/pull/185

hax/treemakers/posrec.py
```diff
@@ -38,11 +38,15 @@ class PositionReconstruction(TreeMaker):
 
        - s1_area_upper_injection_fraction: s1 area fraction near Rn220 injection points (near PMT 131)
        - s1_area_lower_injection_fraction: s1 area fraction near Rn220 injection points (near PMT 243)
+       
+       - s1_range_90p_area: The width of the s1 (ns), duration of region that contains 90% of the area of the peak
```
pdeperio: @skazama Should we put this in ```Extended``` treemaker instead (where ```s1_range_80p_area``` already is)?

hax/hax.ini
```diff
@@ -135,6 +135,12 @@ corrections_definitions = {
         {"run_min": 10223, "run_max": 12089, "correction": "FDC_SR1_data_driven_time_dependent_3d_correction_part3_v1.json.gz"},
         {"run_min": 12090, "correction": "FDC_SR1_data_driven_time_dependent_3d_correction_part4_v1.json.gz"}
     ],
+    "fdc_3d_tfnn": [
```
pdeperio: @jingqiangye Can you add SR0 as well? (Otherwise, I think this will crash on SR0 files.)

@/all When pushing code/analyses from now, for completeness, we should ensure SR0+SR1 are both included.

hax/treemakers/posrec.py
pdeperio: Fixed in https://github.com/XENON1T/hax/pull/190

hax/treemakers/posrec.py
pdeperio: fixed in https://github.com/XENON1T/hax/commit/c994b089d11ed6ab3210b91d32028677dc745796 and https://github.com/XENON1T/hax/pull/174/commits/c1849fd8db8a5d8d7597b2a1abed1f605d6ac9ce

hax/hax.ini
pdeperio: Fixed in https://github.com/XENON1T/hax/pull/174/commits/f60ce1079698a87dfb4cba569438ba73f9a961f8

hax/hax.ini
```diff
@@ -135,6 +135,12 @@ corrections_definitions = {
...
+        {"run_min": 6386, "run_max": 8648, "correction": "FDC_SR1_data_driven_time_dependent_3d_correction_tf_nn_part1_v1.json.gz"},
+        {"run_min": 8649, "run_max": 10976, "correction": "FDC_SR1_data_driven_time_dependent_3d_correction_tf_nn_part2_v1.json.gz"},
+        {"run_min": 10977, "run_max": 13195, "correction": "FDC_SR1_data_driven_time_dependent_3d_correction_tf_nn_part3_v1.json.gz"},
```
pdeperio: @jingqiangye Why does this have different run binning than FANN 3D FDC?

jingqiangye: Because TFNN was using more Kr data for FDC. FANN 3D FDC was using Kr data up to Oct 2017, and TFNN is to Nov 2017. FANN 3D FDC will be updated with latest Kr(Jan. 2, 2018)today.

PR #181, opened 2017-12-18
NN file switching depending on SR

pdeperio:
Add functionality to switch NN files depending on SR number, copied from ```corrections.py``` (this really needs to be refactored soon).

Depends on https://github.com/XENON1T/pax/pull/651.


coderdj:
Can we unmerge this? It's terrible. ;-)

PR #183, opened 2017-12-19
Fix two bugs in new S1 AFT and pattern likelihood calculations

pdeperio:
1. Event index ```i``` was overwritten by a loop index over PMTs, so ```compute_gof``` and ```s1_aft_map.get_value ``` were always using the position of event number 247. Now fixed by using more explicit variable names.

2. The argument ```pmt_selection``` in ```compute_gof``` should be an array of booleans of size ```apc[self.tpc_channels]```. Not sure if what's done in [```BuildInteractions.py```](https://github.com/XENON1T/pax/blob/v6.8.3/pax/plugins/interaction_processing/BuildInteractions.py#L106) is correct, but here it was crashing at 
```
pmts_bottom = np.setdiff1d(self.tpc_channels, confused_s1_channels)
pmts_bottom[self.pax_config['DEFAULT']['channels_top']] = 0

dask.async.IndexError: index 118 is out of bounds for axis 1 with size 118
```
since ```pmts_bottom``` is no longer the full PMT array after removing ```confused_s1_channels```. Now I just explicitly keep a full PMT array ```is_pmt_in```, similarly to [```TopPatternFit.py```](https://github.com/XENON1T/pax/blob/v6.8.3/pax/plugins/posrec/TopPatternFit.py) to keep track of which PMTs to include in the likelihood.

@Requested reviewers, please double check this implementation and implications.


coderdj:
Fine but prefer we test that updates don't break things before merging them. These variables now useless int he production we just did.

PR #184, opened 2017-12-19
Add SR dependent bad PMT list

pdeperio:
Forgot to change bad PMT list depending on SR in https://github.com/XENON1T/hax/pull/181.

Eventually this needs to come together with the NN files.


pdeperio:
Enhanced in https://github.com/XENON1T/hax/commit/c994b089d11ed6ab3210b91d32028677dc745796

PR #185, opened 2017-12-19
Pull corrections logic out of minitree maker

coderdj:
This creates a new class, corrections.py, that handles corrections. Prevents having identical (and copy-pasted) logic in several minitree makers. 

Tested that for one run, cs1, cs2, x_3d_nn, y_3d_nn are identical in previous/this version. 
Also tested for one run that x_observed_nn_tf and y_observed_nn_tf make a hit pattern, s1_pattern_fit_hax vs s1 looks right, and s1_area_fraction_top_probability_hax aren't nans.


pdeperio:
No treemaker version changes, since values shouldn't have changed.

pdeperio:
@ctherreau Please see this PR for how we refactored the corrections functions out of ```CorrectedDoubleS1Scatter.py```

PR #188, opened 2017-12-21
More S1 AFT variables

pdeperio:
New variables recommended by @darrylmasson and @skazama ([note](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenon1t:analysis:sciencerun1:anomalous_background:s1_aft)):

1. ```s1_area_fraction_top_probability_nothresh```: Use area below S1=10 pe instead of switching to hits,

2. ```s1_area_fraction_top_map_value```: The expected S1 AFT value (@darrylmasson, not sure if this is what you meant?)


hax/treemakers/posrec.py
```diff
@@ -23,7 +23,9 @@ class PositionReconstruction(TreeMaker):
        - s1_pattern_fit_bottom_hits_hax: S1 pattern likelihood computed with corrected
                                          position and bottom array hits
 
-       - s1_area_fraction_top_probability_hax: S1 AFT computed with corrected position
+       - s1_area_fraction_top_probability_hax: S1 AFT p-value computed with corrected position
+       - s1_area_fraction_top_probability_nothresh: computed using PE below 10 (instead of hits)
+       - s1_area_fraction_top_map_value: Expected S1 AFT from map
```
darrylmasson: I had intended that we store the value of the binomial pmf (`binom_pmf` from the plugin). Expected S1 AFT isn't hard to figure out (varies pretty closely with z), but the pmf is probably more useful in edge cases.

pdeperio: oh ok. how about now?

PR #189, opened 2017-12-23
Add to HAX new s1 correction from LCE map after electric field effects

MAVJ:
Introduce new correction for s1 when using the LCE map calculated after decouling electric field effects.

- _Changes made in corrections_: 
The new variables is called "cs1_true" (feel free to change the name, I couldn't think about something more catchy)
- _Change made in init_: 
Add the path of the new LCE map for SR1 waiting to be included in PAX, by PR [#658](https://github.com/XENON1T/pax/pull/658)


coderdj:
Have you tested what happens if you try to load a run from SR0? There's no definition of this map for SR0 so will it just ignore the correction or will it fail badly?

Otherwise I might change 'cs1_true' to 'cs1_field_corrected' or something like that. The multiplicative correction factor should also be included in the minitree so we can add/remove single corrections easily (wasn't previously an issue with S1 since we just had the one correction).

Looks good though so we can probably merge quickly once pax 658 deployed.

MAVJ:
Hi @coderdj , 
I didn't test the map up to some moments ago due to lack of time before, but here it is how it looks when used in SR0 data:

corrections_comparison_sr0

Just based on the plot it seems quite similar to the actual correction effects to SR1 data. 

I'm trying to load all Kr83m data from SR0 now to estimate how much time would it take to produce the map for SR0 (as I told you and Pats,  ~1.2 million selected events takes 25-30 min for each Stage of iteration and at the moment we look for at least 7 stages). 

Ideally th new LCE map produced should not depend on energy or field effects, so that it would account only for geometrical effects and therefore it shouldn't have changed that much. But since conditions changed in SR0 I guess that it is better to check if this is true.

pdeperio:
Thanks @MAVJ! So ETA 4 hours for SR0? :P

MAVJ:
xD I just checked how many events after cuts there are and I'm left with 4748307. Meaning 2:00 each stage at least. Should I proceed? I'll start running it and if nothings crashes it could be for tomorrow morning. Let me know what do you think :D

MAVJ:
@pdeperio @coderdj 
Here are the results for SR0 that I run between yesterday/today:

stages_iteration_sr0

3d_map_cy_mean_stage_5_sr0

corrections_comparison_sr0

The results seem to confirm that the maps are really compatible. Here is in any case the LCE for SR0 in case that you want to have both included:

[XENON1T_s1_xyz_lce_true_kr83m_SR0_pax-680_fdc-3d_v0.json.zip](https://github.com/XENON1T/hax/files/1584819/XENON1T_s1_xyz_lce_true_kr83m_SR0_pax-680_fdc-3d_v0.json.zip)

P.S.: I'm sorry I couldn't do more iterations, I got a IOPub message rate exceeded Memory error of some kind

pdeperio:
Thanks @MAVJ! Looks a bit different from previous plot, thus please commit the map to pax and add to here (or if you're already on vacation I can do it). Feliz Navidad

hax/hax.ini
```diff
@@ -144,6 +144,9 @@ corrections_definitions = {
         {"run_min": 0, "correction": "s1-lce_kr83m-sr0_pax-680_fdc-3d_v0.json"},
         {"run_min": 6386, "correction": "s1-lce_kr83m-sr1_pax-680_fdc-3d_v0.json"}
     ],
+    "s1_true_lce_map_nn_fdc_3d": [
+        {"run_min": 6386, "correction": "XENON1T_s1_xyz_lce_true_kr83m_SR1_pax-680_fdc-3d_v0.json"}
```
pdeperio: Is this map using the last iteration? Should it instead be the average of the last few iterations, since it's oscillating?

hax/treemakers/corrections.py
```diff
@@ -11,7 +11,8 @@ class Corrections(TreeMaker):
     Provides:
     - Corrected S1 contains xyz-correction:
       - cs1: The corrected area in pe of the main interaction's S1 using NN 3D FDC
-      - cs1_tpf_2dfdc: Same but for TPF 2D FDC
+      - cs1_true: Same but for after correction of electric field effects
```
pdeperio: If this is what we should all finally use in the analysis, should we call it ```cs1``` instead? (Since previous ```cs1``` is not yet fully corrected, and so we don't need to change the variable in everyone's downstream analyses and ```lax```.)

hax/treemakers/corrections.py
```diff
@@ -11,7 +11,8 @@ class Corrections(TreeMaker):
...
+      - cs1_tpf_2dfdc: Same as cs1 but for TPF 2D FDC
```
pdeperio: Unused? Remove?

hax/hax.ini
```diff
@@ -144,6 +144,9 @@ corrections_definitions = {
         {"run_min": 0, "correction": "s1-lce_kr83m-sr0_pax-680_fdc-3d_v0.json"},
         {"run_min": 6386, "correction": "s1-lce_kr83m-sr1_pax-680_fdc-3d_v0.json"}
     ],
+    "s1_true_lce_map_nn_fdc_3d": [
```
pdeperio: Add SR0

hax/treemakers/corrections.py
feigaodm: Can we keep two so that we can compare? Just in case one has some problem, for eg introducing some anomalous leakage?

hax/hax.ini
MAVJ: Hi Pats,
thanks a lot for making the commit, I wasn't around a pc last days. You are right, the map generated currently comes from the last iteration, since for the inner part of the TPC it seems to be sort of a close to the value of convergance. I could also f¿define after these iterations one extra stage just taking the average value from each FV and generate the map. Would you prefer that?

hax/treemakers/corrections.py
MAVJ: Even though the current cs1 is not the correct one to use, I agree that we could still keep it for comparison (I see how it can be confusing that the variable 'cs1' is not the fully the corrected one, but we can go step by step and update this variable once that the results support the new correction. As for the name of the variable, Dan suggested before to go from 'cs1_true' to 'cs1_field_corrected', both sounds good to me :)

hax/hax.ini
pdeperio: yes, @mcfatelin agree?

(i didn't make any commit?)

hax/treemakers/corrections.py
pdeperio: If we are confident enough (and I think we are from the plots you posted), I think we can directly name the new corrected S1 as "cs1" so all analyses will automatically take the correction, and copy the previous cs1 to "cs1_nofieldcorr" (or something).

feigaodm: @pdeperio Good argument, agreed!

hax/hax.ini
mcfatelin: Sorry I missed this thread. Actually I agree with Patrick that we shall use the average electron lifetime from the last several iterations for extended FV in SR1.

hax/treemakers/corrections.py
MAVJ: Agree. I'll proceed and get the map with the extra suggestions you made and then make the PL for the new maps and also changing the name.

hax/treemakers/corrections.py
MAVJ: I'm not sure if actually someone uses that variables 'cs1_tpf_2dfdc' still for comparison or some other analysis. If it is the case probably it could be removed.

hax/hax.ini
MAVJ: Ciao @pdeperio 
I just wanted to tell you that I just finished the new LCE corrected map introducing your suggestion, this is taking the average (convergence value) for the electron lifetimes in order to define a final Stage 10 correction map. Here are the results:

stages_iteration_calibration_3_feb

This would be the map (for SR1):

[LCE_True_kr83m_stage_10_3d_FDC_t_dep_Calibration_3_Feb_pax-680.json.zip](https://github.com/XENON1T/hax/files/1604876/LCE_True_kr83m_stage_10_3d_FDC_t_dep_Calibration_3_Feb_pax-680.json.zip)

which lead to similar comparison results as before:
corrections_comparison_calibration_3_feb

P.S: Since I'm assuming that we may need too the map for SR0, I'm working already on generating it before making the PR. I'll have it tomorrow morning since this one has to be calculated all again! Let me know if it is what you had in mind and if you have any other suggestion. until tomorrow with the PR.

pdeperio: This looks goods! Thank you!

MAVJ: Here are the results up to Stage 6 for SR0 to generate the LCE corrected map just as done for SR1, I'll now proceed to make the PR:

stages_iteration_sr0

This would be the map (SR0):

[LCE_True_kr83m_stage_6_3d_FDC_t_dep_SR0_pax-680.json.zip](https://github.com/XENON1T/hax/files/1606141/LCE_True_kr83m_stage_6_3d_FDC_t_dep_SR0_pax-680.json.zip)

which lead to the following comparison of results:

corrections_comparison_sr0

PR #199, opened 2018-01-12
Sideband unblinding for AC+Wall model

pdeperio:
This PR unblinds the sideband below the red line and ```cs1 = 200 pe``` in the following image, first defined in https://github.com/XENON1T/hax/pull/168:


feigaodm:
It's too difficult to review it:) Can you try to modify it this way:
lowe_er = "sth"
multiple_scatter = 'largest_other_s2 > 200'
double_e_capture
...
blinding_cut = lowe_er |  multiple_scatter | double_e_capture

pdeperio:
@feigaodm I've implemented your requests. 

Also, just additionally unblinded ```s2<200``` for ```cs1<20``` (```s2<150``` is already unblinded), which I think is what you want as shown in the following figure, right?

blinding_highe

coderdj:
I like the reorganization but considering the confusion we had recently do we want to consider making it more clear that these options shouldn't be touched?

Like, we had more than one analyst (non-maliciously) modifying something called 'blinding_cut'. So that strategy was already a cause for confusion apparently. Now we have a bunch of "unblind_*" options... which at first glance are far more confusing. 

Do you think it's possible to put all the blinding options in a separate section or file clearly delineated as 'do not touch' and not modifiable in override via hax.init?

PR #202, opened 2018-01-17
Add s2_pattern_fit for NN and bugfix s1_pattern_fit

coderdj:
This adds the s2_pattern_fit_tpf and s2_pattern_fit_nn variables to the PositionReconstruction treemaker. Both come from the pax data file (not recalculated like s1_pattern_fit is).

This also fixes a bug in s1_pattern_fit to make it compatible with https://github.com/XENON1T/pax/pull/663

Namely, deactivated PMTs are excluded from the fit. 

Also just noticed my pushes got labeled as Kaixuan. ;-)


skazama:
Is it possible to implement s2_pattern_fit_bottom as Teresa recommended during the collaboration meeting?

coderdj:
In principle it is. In practice we've had trouble getting our pattern fits computed in hax to match those computed in pax exactly. I'm trying to investigate where the problem comes from but haven't been able to yet.

What background should such a cut remove? I'm a bit against adding a bunch of highly construed 99% acceptance cuts to remove single events.

skazama:
If I remember correctly, Teresa suggested to see s2_pattern_fit_bottom instead of s2 maximum hit channel cut.

coderdj:
OK. To be honest I didn't get the point of the s2 pattern fit for the bottom. Like what specific background do we hope to cut with that? I think we should merge what's here (since it has a bug fix and we will have to at least recheck s1 pattern fit) and the s2 pattern fit bottom can be merged in a future PR is we think it's warranted.

PR #205, opened 2018-01-24
TensorFlow bypass

jpienaar13:
This produces PosRec trees successfully for both real and MC data. Avoids the TensorFlow problems experienced on some grid sites.


feigaodm:
I think it's fine to by pass it for now. But probably we will need to use it later if we decide to use it to replace the old FANN.

jpienaar13:
@feigaodm Yes, later we will have to revisit this if/when we replace FANN with TensorFlow. @sthapa is looking at a more permanent fix on that front, however for now we should get MC production back up and running using this workaround.

hax/treemakers/posrec.py
```diff
@@ -55,6 +54,8 @@ class PositionReconstruction(TreeMaker):
 
     def __init__(self):
 
+        #from keras.models import model_from_json
```
tunnell: Why commented?

hax/treemakers/posrec.py
```diff
@@ -90,7 +91,8 @@ def __init__(self):
 
     def load_nn(self):
         """For loading NN files"""
-
+        from keras.models import model_from_json
```
tunnell: It's really sloppy to move imports into packages since machines are setup wrong.  Can you instead look at Mocking, which was the solution I suggested when people were asking about this on Gitter?

JelleAalbers: They are just following our illustrious example at https://github.com/XENON1T/pax/blob/9ecd5c0b94a295893f0bc4eeb0be31039019cd9c/pax/formats.py#L174 :-)

Keep in mind hax.init()  hits all the global imports in treemaker files (since it must figure out what treemakers exist). I don't really want to pull in all the random libraries and files people are using in treemakers -- even if I do have them. I've already been cursing at the corrections treemaker for loading correction maps at import time, now I have to keep pax up to date to even use hax... 

On the other hand if everybody moves their imports inwards dependency tracking becomes a problem. Not sure what wisdom is.

hax/treemakers/posrec.py
pdeperio: @jpienaar13 Can remove if you moved it below.

hax/treemakers/posrec.py
pdeperio: @tunnell This issue is slowing down some analysts on Midway and OSG. Both admins are aware of this, but probably will take a while to get a fix from their side. As @JelleAalbers there are probably a few places already where this would need to be improved (or maybe it's actually the desired method?). Not sure if @jpienaar13 can quickly learn about Mocking, so can we merge this for now?

pdeperio:
@jpienaar13 the new change with ```mock``` looks like it will now be imported on Midway1 again, which was another problem we wanted to solve with this.

i think it was fine before (see my response to Chris above), and seems like just an overall structural choice we need to make.

jpienaar13:
@pdeperio I dont know what the problem was on Midway1. How would this result in it being imported on midway1 but not 2?

Not that I really care which version gets used.

pdeperio:
Sorry, I mean Midway1 is suffering from OS/compiler incompatibility similarly to the OSG sites. So peoples' notebooks fail when they ```import hax``` on Midway1 nodes.

JelleAalbers:
If it's just a missing module, you can do
```
try:
    import keras
except ImportError:
    print("Dude, you don't have keras :-(. Try anything fancy and I'll hit you with 'keras not defined'")
```
and have neither the function-level import nor the mocking (see e.g. https://github.com/XENON1T/pax/blob/master/pax/formats.py#L20).  This assumes the import doesn't cause a segfault, nor succeed while subtly corrupting the essence of reality in a horrible way.

I don't think Mocking + hacking sys.modules is right here. It will cause the code to run happily, but produce nonsense output (e.g. return None where you expected a meaningful value). Hopefully it will still fail, though now later and in a confusing way (NoneType has no attribute blabla..).

pdeperio:
@JelleAalbers it's not missing the module, it's a runtime error:
```
$ haxer --main_data_paths /project2/lgrandi/xenon1t/processed/pax_v6.8.0 --input 170925_0622 --pax_version_policy v6.8.0 --treemakers Basics --event 1

ImportError: /lib64/libc.so.6: version `GLIBC_2.16' not found (required by /project/lgrandi/anaconda3/envs/pax_v6.8.0/lib/python3.4/site-packages/tensorflow/python/_pywrap_tensorflow_internal.so)

Failed to load the native TensorFlow runtime.
```

JelleAalbers:
OK, then I'd just go for the function-level import. You could do a generic try-except but I doubt it's worth it (it might stay in the code then fail to alert us to an important problem later).

jpienaar13:
So the consensus is that I should revert to the function level import?

pdeperio:
Yes please. @tunnell approve?

tunnell:
I did give @jpienaar13 the psuedo-code to do this in hax.init(). If he can, that would be the best way to do it.  Pass a mc=True to hax init, then catch that and mock within hax.init.  Then you don't need to change the tensorflow/keras import to within a function.   I'm approving it, but please try for 30 minutes to see if the code I sent would work.  Otherwise, just merge.

pdeperio:
This is not just a MC issue as I mentioned [above](https://github.com/XENON1T/hax/pull/205#issuecomment-361359625).

jpienaar13:
@tunnell I did that. Look at https://github.com/XENON1T/hax/pull/205/commits/29b50b7d7e2d278e679a9b5238c9714eff51f0b4

tunnell:
It would work if you renamed MC to just disable-TF

tunnell:
If TensorFlow broken on OSG, then no fix in hax will help.  Right? 

Yes, I'm saying MC flag and also midway-login1 check.  We should move JupyterHub to 2 or stop using it.

PR #206, opened 2018-01-30
Stricter blinding cut

pdeperio:
Implement a more stricter blinding cut (as requested [here](https://github.com/XENON1T/hax/pull/199#issuecomment-360462399)):

1. Move definition from ```hax.ini``` into separate ```unblinding.ini``` file (the new name should be more scary to edit).

2. Move cut string from ```hax.config``` to ```hax.unblinding_config``` such that it cannot be modified with ```hax.init(kwargs)``` as before.
  a. Thus, https://github.com/XENON1T/hax/pull/204 becomes obsolete since it's not allowed at all here. @JelleAalbers what other legitimate uses were you thinking?

3. Rename ```blinding_cut``` to ```unblinding_selection``` to break backwards compatibility of any rogue scripts (though should already be taken care of by 2. above)

4. Automatically unblind MC (previously, manually using ```kwargs``` in secret MC notebooks and [laxer](https://github.com/XENON1T/lax/blob/v1.3.0/bin/laxer#L126)).
  a. Moved ```mc_data``` variable from ```minitrees.py``` to ```hax``` global variable to reduce code duplication.
  b. Reduced code duplication for MC checks and e-lifetime.

Henceforth, the only way to unblind is to submit a PR modifying ```unblinding.ini``` (or still a few other previous malicious ways, but please don't).


pdeperio:
Added the SR0 plot [here](https://github.com/XENON1T/hax/pull/168#issuecomment-361419598). Quite similar to SR1 above, so I don't think using the same lines for SR0 will bias us.

coderdj:
OK. Might want a second or third review also since important.

hax/__init__.py
```diff
@@ -23,6 +24,13 @@
 # Just do import hax, then hax.init(), then access config has hax.config.
 config = {}
 
+# This stores the unblinding selection (do not try to modify directly; unblinding.ini
+# file should be modified with proper PR and review if needed)
+unblinding_config = {}
+
+# Flag for whether you are working with MC (True) or not
+mc_data = False
```
JelleAalbers: Could you clarify what this global mc_data is doing? It looks like this is a run-level property, but it's stored as a package global. You might get some interesting behaviour if you're doing real data / MC comparison studies.  Could the treemakers that need to know whether they are in MC land (or maybe event the treemaker base class) call hax.runs.is_mc instead?

hax/__init__.py
```diff
@@ -63,6 +72,19 @@ def init(filename=None, **kwargs):
...
+
+    global unblinding_config
+    unblinding_configp = ConfigParser(inline_comment_prefixes='#', strict=True)
+    unblinding_configp.read(unblinding_filename)
+
+    # Evaluate the values in the ini file
+    unblinding_config = {}
+    section_to_use = 'UNBLINDING'
+    for key, value in unblinding_configp[section_to_use].items():
+        unblinding_config[key] = ast.literal_eval(value)
```
JelleAalbers: Since you're only storing two settings  in unblinding.ini, and the unblinding.ini filename can't be overriden by the user, would it be easier to make unblinding.ini just unblinding.py? Then add unblinding here https://github.com/XENON1T/hax/blob/master/hax/__init__.py#L7, and you can access the settings as `hax.unblinding.unblinding_selection` and `hax.unblinding.blind_from_run` when you need them in hax.runs.

hax/corrections_handler.py
```diff
@@ -116,10 +116,21 @@ def get_correction(self, correction_name, run, ismap=False):
...
+        if hax.mc_data:
+            elifetime = self.get_misc_correction("mc_electron_lifetime_liquid", run_number)
+
+        else:
+            try:
+                elifetime = self.get_electron_lifetime(run_start, value)
+
+            except Exception as e:
+                print(e)
+                return 1.
```
JelleAalbers: If there is no electron lifetime available, are you sure you want to produce a correction minitree? 

The generic except catching with a default of 1 instead of NaN might cause some fun later on. If someone puts a syntax error in `get_electron_lifetime`, everything will run happily but we'll get no electron lifetime corrections. Hopefully someone is checking the logs and they'll see there was a syntaxerror somewhere (but since it's just a print they also don't know where).

hax/runs.py
```diff
@@ -307,6 +307,16 @@ def get_run_start(run_id):
         # return np.datetime64('2017-06-13T18:17:43.000000000')
 
 
+def is_mc(run_id):
```
JelleAalbers: Glad you cleaned up this code duplication!

hax/__init__.py
pdeperio: it will soon have many entries after #199 [here](https://github.com/XENON1T/hax/pull/199/files#diff-f5b19e12c8f52ab242b67b1d087318a1R114). but i guess your suggestion applies there too, and would clean up the ```init``` function, right?

hax/corrections_handler.py
pdeperio: I think @coderdj implemented this at first; @coderdj what was your motivation for this?

hax/__init__.py
pdeperio: Remnant from confusion during development. Fixed in 1154d04.

hax/__init__.py
JelleAalbers: Yes, it should work just as well. I haven't looked at #199 in detail, but maybe you want to put all settings in a `settings` dictionary inside `unblinding.py` if you have to do a lot of dynamic key access (`getattr(hax.unblinding, key)` isn't so great). You could even move `is_blind` into the file as well to have all blinding logic together.

hax/corrections_handler.py
coderdj: I copy-pasted code from Jacques who had added this condition for MC I think (or at least that's what I thought when I copy-pasted). But we could remove it as far as I'm concerned.

hax/__init__.py
pdeperio: Great suggestion! Implemented in 1cc8c65 and 309112f

hax/corrections_handler.py
pdeperio: OK raising the exception instead in e3525fd

jpienaar13: @coderdj Pretty sure I didnt add the e-lifetime=1 option. in fact i believe git blame says you did.

hax/__init__.py
```diff
@@ -1,11 +1,12 @@
 import logging
 import os
+import ast
```
JelleAalbers: This is now unused, though I suppose you could replace the eval of the main config with this (currently it has a "slightly" less secure eval with the `os` module exposed (which is also no longer used))

hax/corrections_handler.py
```diff
@@ -116,10 +116,20 @@ def get_correction(self, correction_name, run, ismap=False):
...
-    def get_electron_lifetime_correction(self, run_start, drift_time, value='DEFAULT'):
+    def get_electron_lifetime_correction(self, run_number, run_start, drift_time, mc_data, value='DEFAULT'):
         """Wrapper that does the exponential calculation for you
         """
-        return np.exp((drift_time / 1e3) / self.get_electron_lifetime(run_start, value))
+        if mc_data:
+            elifetime = self.get_misc_correction("mc_electron_lifetime_liquid", run_number)
+
+        else:
+            try:
```
JelleAalbers: You can remove the whole try-except now, it just raises every exception it encounters (which happens anyway)

pdeperio: Fixed in 01169a1

hax/__init__.py
pdeperio: Just removed in 6ad82f6 (since I couldn't easily just swap ```eval``` with ```ast.literal_eval```).

hax/corrections_handler.py
coderdj: Yeah now that I look at it I did. I don't know why I did.

But the other part is true. When I did the refactor of the corrections stuff I also noticed it was weird and copy-pasted it thinking you added it for MC stuff (without checking if that was the case). 

Anyway, it can be removed.

PR #214, opened 2018-02-14
Update peak_treemakers.py

katrinamiller:
Update peak_treemakers.py with the SingleElectrons minitree class for SE waveform shape analysis.


tunnell:
Can merge after tests pass

feigaodm:
Hi @katrinamiller , thanks for the PR. However, I don't think it's so necessary to add this SingleElectron Treemaker since we have this Isolated Peak there. For eg, I used IsolatedPeak treemaker to do SingleElectron Analysis, see [this notebook](https://github.com/XENON1T/SR1Results/blob/master/Background/AccidentalCoincidence/se_classification.ipynb), you can simply use this hit bounds stuff in the pre-selection.

tunnell:
It uses about half the disk space (i.e., 50 GB total).  For some reason, applying the cuts after the fact to try to generate the equivalent of the SingleElectron selection doesn't give the same number of peaks as if you just ran the tree maker.  Therefore, it would be useful to run this on everything to understand more what was going on.

@feigaodm did you check that you would get the same number of peaks as if you just ran the treemaker on its own?

tunnell:
(Added Fei so he has to approve too)

tunnell:
@katrinamiller can you explain a little about the issue you'd be having to fill @feigaodm in on the background.

feigaodm:
haha, no, I haven't check it myself. Original I was using the SingleElectrons Treemaker from @katrinamiller , Then I switched to this Isolated peaks since it's also very convenient. It would be nice if she can confirm if it's necessary to have this added. I don't have a strong preference :)

tunnell:
It's not 'necessary' but convenient (for analysts and debugging this) since only 50 GB.  Unless it's annoying for processing?  I'll ask Katrina to write a few sentences then you can comment.

katrinamiller:
hi @feigaodm - the issue i'm having is loading the dataset from the hax.paxroot.loop_over_dataset command produces a different number of peaks then when I load the minitree from hax.minitrees.load & applying the cuts externally. I need the dataframes to be the same to complete the SE waveform shape analysis.

feigaodm:
@katrinamiller Thanks for explanation but I don't understand your problem. I will just approve and @tunnell can merge it.

PR #215, opened 2018-02-15
Add DEC unblinding for 14 days after AmBe data

cwittweg:
Unblind runs 8340-8728 from 50-80 keV to assess the removal of I-125 after activation via AmBe. Corresponding note can be found here: https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenon1t:analysis:sciencerun0:check_iodine125_removal_in_ambe_data


hax/unblinding.py
```diff
@@ -57,6 +57,9 @@
 unblind['dec'] = {}
 unblind['dec']['e_range'] = '((0.0137*(cs1/.1429 + cs2_bottom/11.36) < 50.) | (0.0137*(cs1/.1429 + cs2_bottom/11.36) > 80.))'
 
+# Unblinding of 14 days post-AmBe data for I-125 removal assessment
```
pdeperio: Please add this PR number (i.e. "see #215") to the comment for future reference.

hax/unblinding.py
```diff
@@ -57,6 +57,9 @@
...
+unblind['dec']['e_range_i125'] = '((run_number>=8340) & (run_number<=8728) & (0.0137*(cs1/.1429 + cs2_bottom/11.36) > 0.))'
```
pdeperio: I don't think ```E > 0``` term is necessary?

tunnell:
Technically fine.  Just note that you shouldn't use these 15 days later in your search.  Also, 15 days isn't enough to say anything interesting about DEC, right?

PR #216, opened 2018-02-24
SR0+SR1 Unblinding proposal

pdeperio:
For [SR0+SR1 unblinding](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenon1t:analysis:sciencerun1:unblinding_proposal) cs1 < 80 pe.

(Also, consolidating some redundant strings.)


coderdj:
Agreed to put cs1 to zero?

PR #217, opened 2018-02-27
Standard set of minitrees for Science Run available at hax.load()

tunnell:
Most use cases load a standard set of minitrees for use with lax for most single scatter analysis.  This means that now we can do:

`hax.minitrees.load(28192, minitrees='all')`

The list of minitrees came from:

https://github.com/XENON1T/SR1Results/blob/master/Calibration/inputs_for_bbf/reduce_data.py#L137


tunnell:
Do people want 'all' to be default?  That would be a reasonable default but I don't want to break compatibility with existing code.

darrylmasson:
The default 'Basics' and 'Fundamentals' used to cover a lot, but without (x,y,z) and cs1 and cs2, I don't think you can do much with them anymore, so I'd be surprised if any existing code relying on the defaults still worked anyway.

tunnell:
@darrylmasson fixed, default to all

tunnell:
Can somebody not me merge before this goes stale?

PR #218, opened 2018-03-03
Unblind energy sidebands in cs1

coderdj:
Once we're ready this should open up cs1<3 pe as well as 70<=cs1<80 pe. Unblinding low and high energy sidebands for WIMP search. 

It's very debatable whether the high energy sideband is needed at all and I can't really think of a justification at the moment so we should discuss.


coderdj:
👍 (I can't approve since I created)

feigaodm:
What do we do if we see 10 events in 70-80 PE range?

pdeperio:
We are just doing ```cs1 < 3``` right now.

feigaodm:
good, I didn't realize this new change. So 70-80 is not nessary at all.

Since this region won't be analyzed directly, do you think we can cut this if background model can not explain there at all? In my opinion it's ok as long as we don't cut anything below 70.

feigaodm:
not necessary anymore PR #216

PR #219, opened 2018-03-05
Check for metadata existence in ROOT file

pdeperio:
Sometimes metadata doesn't get saved in ROOT file for some reason. So raise error if so.


tunnell:
Can it just be a warning?

pdeperio:
Previously this would stop people's notebooks, without any clear debugging info (also hard to find during cax auto-processing). Then there would be a lot of digging and eventually we find the corrupt minitree, delete it, and regrow. This PR is so we have that corrupt file info immediately so it can be more quickly fixed (either picked up by cax or user report). Shouldn't just warn and continue analysis on this file in case something is wrong with it.

tunnell:
fair enough

PR #211, opened 2018-02-03
PosRec minitree unblinding

pdeperio:
The ```PositionReconstruction``` minitrees (ROOT files on disk) are mistakenly blinded due to [this line](https://github.com/XENON1T/hax/blob/v2.3.3/hax/treemakers/posrec.py#L122). @feigaodm noticed after the recent ```s2 < 200``` unblinding attempt (https://github.com/XENON1T/hax/pull/208), all the ```PositionReconstruction``` variables of events in the previously blinded region were ```NaN```.

To not stop the analysis, I did something naughty and commented out the blinding in ```minitrees.py``` [here](https://github.com/XENON1T/hax/blob/v2.3.3/hax/minitrees.py#L404) in my local installation:
```
    #if (hax.unblinding.unblinding_selection not in preselection and
    #    ('Corrections' in treemakers or
    #     hax.treemakers.corrections.Corrections in treemakers) and
    #        hax.unblinding.is_blind(run_id)):
    #    preselection = [hax.unblinding.unblinding_selection] + preselection
```
and regenerated the minitrees with version ```1.2``` as of today.

Do not merge this yet, in case someone is using some non-tagged runs that haven't been regenerated by cax yet, and end up regenerating themselves with the blinded version. We need to figure out how to fix [that line](https://github.com/XENON1T/hax/blob/v2.3.3/hax/treemakers/posrec.py#L122) and include in this PR.

PR #220, opened 2018-03-18
Unblind the DEC region after NG data after SR1

alexfieg:
In order to investigate the I-125 removal further we need to unblind the region between 50 keV and 80 keV in the data past the NG run beyond SR1. We will not use this data anyway for the DEC search as we fixed ourselves to SR1.


pdeperio:
Add [link to note](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenon1t:analysis:sciencerun0:check_iodine125_removal_in_ambe_data#projection_for_neutron_generator_sensitivity_to_higher_removal_efficiencies)

coderdj:
Can merge. Shingo brought up the valid point that the E lifetime is wrong right now to blinding boxes in incorrect location.

feigaodm:
that's true, but when producing these minitrees in the correction, they should have correctd S1 and S2s in the correct way. I think that will fix this problem.

skazama:
I don't know correct e-lifetime model is already implemented in hax or not. We had a huge e-lifetime drop last month, which is not modelled currently, so I think even if we produce correction minitree by yourself, cs2 is currently wrong.

feigaodm:
By they I mean @alexfieg and @cwittweg :)

alexfieg:
We are aware of this and Michelle and the shifters had a look into this. But we will correct this manually using other lines and the upcoming Kr83m measurement

cwittweg:
But without e-lifetime model we cannot guarantee that the right energy region is unblinded. We should update the model which might be possible even without Kr-83m calibration by using the activated xenon lines during and after the NG run. These are not blind and comparable to Kr-83m. I can look into this once I am back from DPG.

cwittweg:
Okay, the way the blinding cut is defined this should not be an issue, but we should update the e-lifetime model anyhow.

feigaodm:
I think it's safe to keep this data unblinded as this won't be used to search for new physics especially at this energy range. Be careful about the correction and energy re-calibration, otherwise you won't be able to see your expected signal easily.

pdeperio:
Hang on... sorry, I just realized https://github.com/XENON1T/hax/pull/211 might affect you now (PosRec minitrees will not be simply unblinded with this PR, and need to regenerate them...)

feigaodm:
yeah, that might had affect their analysis after AmBe too, right?

pdeperio:
The previous post-AmBe unblinding should be fine, their ```PositionReconstruction``` minitree version is ```1.2```. But latest datasets are ```1.1```, which is bad, please wait while I work with @ershockley to get this fixed.

PR #186, opened 2017-12-20
New data map names

l-althueser:
This PR changes the data map names to the ones in https://github.com/XENON1T/pax/pull/654
Wait for this PR in PAX first or cancel both ;)

PR #222, opened 2018-03-26
fix S1AFT map name

feigaodm:
In PR #186, S1 AFT map name is not updated. Thanks @jpienaar13 for reporting. 

In the meantime, I fixed the tensorflow filename for bug fix. Should work only if https://github.com/XENON1T/pax/pull/679 is merged.


feigaodm:
thanks for quick check @l-althueser

PR #223, opened 2018-04-03
Create a license file

tunnell:
GPLv3 license file reflecting what was written in README

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,14 @@
+Copyright (C) 2017  Jelle Aalbers and Christopher Tunnell
+
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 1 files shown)


JelleAalbers:
Good to add a license, guess we never got around to it... We usually did MIT or BSD, any reason to add copyleft specifically to hax? I guess it protects against someone trying to take hax off the air in the future, but it might make it harder for other experiments to adopt hax in a closed-source copy someday (there is stuff like database usernames and urls in the config).

tunnell:
My current thinking is Apache 2 instead of BSD for things that could be private (it's more explicit about citing the work you used), LGPL3 for libraries, and GPL3 for public things.  We just need to make sure that the config is separate from the tool I think.  Every package in XENON has a different license at the moment though :)  Also, everything tested, maintained, or abandoned differently... trying to figure out what we have to do for nT.

tunnell:
@JelleAalbers you agree?  If so, approve?  Otherwise, discuss?

JelleAalbers:
I'd be more comfortable starting with a non-copyleft license. There's a good case for copyleft, but also a case against it; and once we add it, it can never be removed. At least I'd like to hear some more thoughts on this before we proceed.

tunnell:
Apache2?

Issue #179, opened 2017-12-13
hax.cuts.apply_lichen reports lichen version as nan for some cuts


miwig:
apply_lichen [tries to get the lichen version from the uninstantiated class](https://github.com/XENON1T/hax/blob/39c82e1eb3bc64a88180a834268be8d95dcaa30f/hax/cuts.py#L227), but some lichens, like `S2AreaFractionTop` [only set their version when instantiated](https://github.com/XENON1T/lax/blob/c7d781de5e3f3c7df4b1595c31ac0e8a5301e98d/lax/lichens/sciencerun0.py#L540).

To reproduce:  
```
data = cuts.apply_lichen(data,'S2AreaFractionTop')
cuts.history(data)
```

pdeperio:
@adambrown1 do we still need the 2 versions right now? or how about splitting it off into another cut like ```S2SingleScatterSimple``` does?

adambrown1:
We probably no longer need two versions. But I think there is also a
trivial fix by moving the first pair of parentheses () from line 223 in
cuts.py to the end of line 220. I can make a pull request tomorrow or you
can change it now.


2017-12-13 17:58 GMT+01:00 Patrick de Perio <notifications@github.com>:

JelleAalbers:
This should be fixed in #224

PR #224, opened 2018-04-03
New tutorials

JelleAalbers:
This adds a trilogy of up-to-date tutorial/howto notebooks for hax. The first two are targeted at new students while the last one may be useful to more seasoned analysts.

I moved the old tutorials into an 'old' folder. All the ones I looked at would no longer work; most of them still reference XENON100 or features that no longer exist or have changed incompatibly. However, some of them are still valuable as documentation for features not described elsewhere. It will take some time to sort these out -- of course input is welcome here.

This also adds a few small things I found useful while writing them:
  * `hax.cuts.apply_lichen` now accepts a `deep=True` option. This will go one level down the rabbithole of lichens, so you get an elaborate cut report even if you just use a summary lichen such as LowEnergyBackground. See the second tutorial for an example. Also includes a fix for #179.
  * `hax.runs.get_run_info` now accepts pandas series and numpy arrays of run number/names, rather than just lists and tuples.
  * Added a small helper function `hax.runs.tags_count` that counts how often a tag occurs in a selected dataframe of datasets. Again an example is in the second tutorial.

After this is merged, we can advertise the tutorials on the README page and in the analysis guide on the wiki.


tunnell:
I flipped through the examples.  They look really nice and informative.  I'd just delete the old ones (it's why we have version control) since I think they're more confusing than not.  We also have SR1Results and other repositories that people can copy-paste from.

PR #226, opened 2018-04-30
Unblind sidebands for MC Matching (DEC related)

alexfieg:
In order to clarify our matching on Pb-214, we need a region which is clean beside the low-E region. This is only given between 50 keV and ~ 75 keV. The chosen further unblinding from 50 keV to 58 keV and 70 keV to 80 keV would be very useful and is still more than 4 sigma away from the peak center ( 64.33 keV with sigma/E ~ 2.6 keV, according to the resolution from: https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenon1t:g1g2resfordec).

Additionally, we updated for the DEC region the g1 and g2 values (should not change much) as used in SR1. Did not do it for double beta decay as I do not want to interfere with other analysis of anyone.


alexfieg:
changed to 3-sigma as Fei pointed out the difference between one-sided and two-sided (my bad)

cwittweg:
Can we go ahead with this?

PR #235, opened 2018-08-01
Blinding update 0vbb

feigaodm:
This is to narrow down the 0vbb blinding region to Q±4sigma to study the background peaks better, the energy is defined in a [recent note from Chiara and Tianyu](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenon1t:chiara:double_beta:update_pax6.10.0)

Validation plots (not informative because there is no cuts) to be removed after merging. 
blinding


hax/unblinding.py
```diff
@@ -66,7 +66,7 @@
 
 # 0vbb blinded from 2.3-2.6 MeV (see #161)
 unblind['0vbb'] = {}
-unblind['0vbb']['e_range'] = '((0.0137*(cs1/.1429 + cs2_bottom/11.36) < 2300.) | (0.0137*(cs1/.1429 + cs2_bottom/11.36) > 2600.))'
+unblind['0vbb']['e_range'] = '((0.0137*(cs1/(z*0.0001 + 0.1463) + cs2_bottom/(-0.017*z + 10.628)) < 2457.83*(1 - 4*0.01)) | (0.0137*(cs1/(z*0.0001+0.1463) + cs2_bottom/(-0.017*z +10.628)) > 2457.83*(1 + 4*0.01)))'
```
tunnell: __[Flake8]__

[E501] line too long (213 > 120 characters)

sanderbreur:
Shouldn't we again do a sort of side-unblinding? A short note on what the
expected rates are before we unblind those regions? We have seen with the
DEC analysis, that even we really though we understood it, doing it with an
unblinding proposal helped us gain a lot more understanding.

Op wo 1 aug. 2018 om 15:04 schreef Patrick de Perio <
notifications@github.com>:

feigaodm:
@sanderbreur Spectrum is just 1 day of test data, of course we didn't open everything all together :)
Doing a sideband unblinding requires some knowledge about BG modeling, but this is just to make the blinding cut more reasonable.

Issue #232, opened 2018-06-11
when loading data: ValueError: Metadata mismatch found in `from_delayed`


NataschaRupp:
Hi,
I would like to load data after SR1 (in the example Feb 1, 2018 - Feb 3, 2018). It seems like there is a mismatched between expected data type and found data type. I attached a notebook where one can reproduce the error and the error message itself. I run on midway2-login2.
 
[hax_issue.ipynb.zip](https://github.com/XENON1T/hax/files/2089794/hax_issue.ipynb.zip)

[error_message.pdf](https://github.com/XENON1T/hax/files/2089790/error_message.pdf)

Thanks, Natascha

LeafSwordy:
Hi,
I got a similar error, but I am using pax v6.8.0.

[hax_issue_6.8.0.ipynb.zip](https://github.com/XENON1T/hax/files/2122990/hax_issue_6.8.0.ipynb.zip)
[metadata_error.txt](https://github.com/XENON1T/hax/files/2122991/metadata_error.txt)

pdeperio:
Can you tell which run is failing? (I guess run ```16516```, which has a very small number of events. @JelleAalbers do you think https://github.com/XENON1T/hax/issues/47 is related?)

AlexElykov:
Hi, 

@pdeperio 

Encountered the same issue as above while loading NG data, caused by Proximity minitree. Can try to give more details if needed.

MAVJ:
Hi @pdeperio, 
Just that somebody knows, I went through the whole background data in SR1 and found similar errors for the following runs:

6932, 6945, 7148, 7178, 7181, 9383, 9398, 10579, 11019, 12736, 13423, 13554

JelleAalbers:
This should be solved by #236. If you still see this in recent hax versions, let me know, and I can have a look.

PR #236, opened 2018-08-03
Force df type match during minitree load

JelleAalbers:
This is a workaround for #232: it forces all minitrees into the same format during loading, set by the first dataset given to the load command.

The columns and data types a treemaker produces are determined dynamically from the output. Thus, inevitably, some minitrees produce a different set / type of columns in rare situations. There is also the mysterious 'index' column, which appears under unknown (to me) circumstances in pandas and can find its way into a minitree. 

When either of these happens, current versions of dask throw the 'metadata mismatch' error. Since the errors started to appear recently, I assume previous versions did something else (perhaps pathological).

After this PR, if you load 1000 minitrees and number 523 has one of the defects above, it will be adapted to fit the format of the first minitree. Missing columns are filled with NaNs (INT_NAN from pax in the case of integer columns), and extra columns are deleted. Type mismatches are resolved as best as we can too. 

If the first minitree has e.g. a spurious index column (the most common problem reported in #232), you would now get an extra 'index' column for the entire set of runs you're loading (but for most runs it will be set to NaN).

This is a workaround: a proper fix would be to change the minitree format to require a full specification of the data type of every minitree, and enforce it in hax. This is the approach of strax.


JelleAalbers:
Yes, you can reproduce the problem (#232) with
```
import hax
hax.init(minitree_paths=['/project/lgrandi/xenon1t/minitrees/pax_v6.8.0',
                         '/project2/lgrandi/xenon1t/minitrees/pax_v6.8.0'],
         pax_version_policy='6.8.0')
df = hax.minitrees.load([6931, 6932])
```
and verify it is fixed in this branch. (thanks to Miguel in https://github.com/XENON1T/hax/issues/232#issuecomment-408857357 for listing the problematic run numbers)

feigaodm:
@JelleAalbers Does it change total number of good events in the whole SR1 dataset? It would be interesting to see if some good events are removed by this. But it looks fine to me.

feigaodm:
@JelleAalbers Thanks! By your explanations I'm confident enough that this won't be a big issue, but agree to backup. Maybe a new copy won't be necessary. But I would think to make a copy of minitrees to a folder like pax_v6.8.0_hax_v2.4.0. Do you agree @pdeperio ?

JelleAalbers:
Sorry if I wasn't clear, this won't change any minitrees you already made, or even minitrees you're making in the future. It just changes how they're loaded, specifically, how minitrees from different runs are combined. Yyou probably do want to make a backup of the cache file / h5 / csv you got for the final dataset, but I assume you already have that.

feigaodm:
@JelleAalbers Yes, I know this detail. Maybe I was trying to make it more general. In principle we shall make the copy anyway. I think we shall make hax_v2.5.0 after this PR is merged.

JelleAalbers:
OK, sounds good!

pdeperio:
OK I've started the copy to ```dali-login.rcc.uchicago.edu```:
```
cp -r /project*/lgrandi/xenon1t/minitrees/pax_v6.8.0 /dali/lgrandi/xenon1t/minitrees/.
```

For reference, the data post-lax is stored [here](https://github.com/XENON1T/SR1Results/tree/PRL_v1/StatisticalAnalyses/inputs_for_likelihood/lax_1.5.1_egg/data) and the README contained therein.

PR #238, opened 2018-08-19
Bypass lax for applying lichens on empty dataframes

JelleAalbers:
This changes hax.cuts.apply_lichen to skip the lichen evaluation when the dataframe is empty.

Lax's lichens crash when run on empty dataframes. If you do e.g. `lax.lichens.sciencerun1.AllEnergy().process(df.iloc[:0])`, where df is any normal dataframe, you will get an error like this: `ValueError: cannot set a frame with no defined index and a scalar`. The problem seems to come from setting new columns on dataframes (in certain ways?) -- which lax, and individual lichens, do in several places. 

 Empty dataframes arise in practical usage in two ways:
  * You're applying several lichens sequentially, and all events in a dataframe are cut before you get to the last one;
  * You're loading many datasets with `make_minitrees = False` and lichen preselections (see #229). If a minitree is missing, `make_minitrees = False` causes hax to return an empty dataframe for that run.  Since lichen preselections are evaluated run by run, this means lichens get an empty dataframe.

It might be possible to fix this on lax' end, however:
  * It would entail some digging through and modifying various (frozen) lichens, with all the risks that entails;
  * There is no need to even call lax if the dataframe is already empty;
  * Lax was not made as a tool for applying cuts: it's main function is annotating dataframes with which events pass which cuts. In Chris' original workflow, you would load all data into RAM, run all lichens, get many CutXXX columns, and finally apply all cuts at once (or a subset, if you like). In this case, you would not hit this bug.


hax/cuts.py
```diff
@@ -246,18 +246,22 @@ def apply_lichen(data, lichen_names, lichen_file='postsr1', deep=False, **kwargs
...
-                             getattr(d, 'Cut' + l_name),
-                             desc=desc,
-                             **kwargs)
+            if len(data):
+                # .copy() to prevent pandas warning and pollution with new columns
+                d = l.process(data.copy())
+                bools = getattr(d, 'Cut' + l_name)
+            else:
+                # Some (all?) lichens crash on empty dataframes, which can happen
+                # if make_minitrees=False.
```
tunnell: __[Flake8]__

[W291] trailing whitespace

feigaodm:
Thanks @JelleAalbers ! It looks nice from my side. I'll merge it unless there is objections or planned change

PR #242, opened 2018-11-14
Add LoneS2 and BigBeforeSmall treemakers

JelleAalbers:
This adds a treemaker for producing information on lone S2 signals, as well as a separate treemaker that extracts basic info about all large S2s close to very-small-S2 events.

See also [this note](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenon1t:s2only:ubp:partial_unblinding_check) and other notes linked on our [general overview page](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenon1t:s2only).

I also included a tiny fix to `minitree_formats.py`. In the code that checks if array fields occur, we currently raise an error when an empty dataframe is encountered. However are legitimate cases (the BigBeforeSmall treemaker is one of them) where a treemaker produces empty dataframes, and we don't want to crash a large minitree load just for this one check. Thus, I changed the behaviour of the array field check on empty dataframes to printing a warning message (and assuming there are no array fields).


hax/treemakers/lones2info.py
```diff
@@ -0,0 +1,228 @@
...
+        largest_s2_area = s2_areas.max()
+        s2s = [p
+               for p in s2s
+               if p.area > max(300, largest_s2_area * 0.1)][:5]
+
+        result = []
+        for i, p in enumerate(s2s):
+            for rp in p.reconstructed_positions:
+                if rp.algorithm == 'PosRecTopPatternFit':
+                    x, y = rp.x, rp.y
```
feigaodm: I think it's necessary to add the neural net pos which is the "official choice" in SR1 paper. And probably tensorflow position is helpful for some purpose.

hax/treemakers/lones2info.py
```diff
@@ -0,0 +1,228 @@
...
+        #   area > max(300 PE, 0.1 * largest_s2_area)
+        s2s = [p
+               for p in event.peaks
+               if p.type == 's2']
+        if not len(s2s):
+            # Hm, that's somewhat interesting...
+            # Maybe have a look at these sometime.
+            return []
+
+        s2_areas = np.array([p.area for p in s2s])
```
feigaodm: Can we add x, y corrected S2 for top and bottom array so that we can use cS2Asymmetry for cuts? I guess it's done in analysis notebook if not here, then it's fine. But putting it here might be nicer to reduce complexity in analysis scripts.

hax/treemakers/lones2info.py
JelleAalbers: Good idea: I've added the neural net position. We've been using TPF so far, but we should be able to compare them. 

From https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenon1t:analysis:sr1:posrec#why_nn, it sounds like NN was favoured for better performance near dead PMTs and smoother overall response. In contrast TPF should work better on very low S2-events, which is the most important for us.

hax/treemakers/lones2info.py
JelleAalbers: Yes, we use the (x,y)-corrected AFT for the AFT cut, which we indeed compute in the framework. The computation is done automatically whenever the AFT cut is activated (see https://github.com/XENON1T/s2only/blob/master/s2only/utils.py#L129).

feigaodm: ok, it's fine but another analyzer will need to implement it him/herself

hax/treemakers/lones2info.py
feigaodm: I think we are (at least I'm) missing a plot to support the argument that TPF works better in the lowest energy, though this has nothing to do with this PR.

PR #243, opened 2018-12-13
new TFNN with PMT31 off

weiyuehuan:
Update TFNN with PMT31 off after run: 22130

Reference: https://github.com/XENON1T/pax/pull/721


weiyuehuan:
@jnaganoma @mcfatelin @feigaodm Please check the min run number, thx.

weiyuehuan:
@feigaodm @mcfatelin we have run dependent TFNN setting in hax level like this pull, but I do not remember if we have it in pax level for FANN? ..

skazama:
@weiyuehuan @mcfatelin PMT 31 was turned off at Sep 11th, but its gain was really low (<< 1e6) after around Apr 1st, so we might need to exclude this PMT even before Sep 11th, depending on the estimated SPE acceptance. Currently, this PMT is already excluded in the gain model after Apr 1st.

feigaodm:
I agree with @skazama , can @weiyuehuan update the PR if there is no objections?

PR #244, opened 2019-02-13
Blinding the region below 10keV first in SR2

mcfatelin:
This PR is for blinding the region <10keV in SR2. Future unblinding of control sample to study the cut behavior will follow up soon.


mcfatelin:
> I am wondering if we should use the z-dependent g1/g2 values here while the low-E ER group does not take the z-dependence into account. It might be confusing for the analysts if the spectrum in CES does not stop exactly at 10 keV. Furthermore, I think the numbers used here were evaluated with pax 6.10.1 (https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenon1t:chiara:resolution_6.10.1) while the ones for pax 6.8.0 should be used (https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenon1t:analysis:backbround_model_dec_sr1).

In my sense, the purpose of blinding is to prevent analysts looking into data bias their analyses (looking for what they want). In low energy ER, then it is the excess. I don't think fine-tuning the g1 and g2 are necessary since the small number of events leaked into "blinding region" because of the distinct g1/g2 used won't give any info about the excess here.

And I thought we're going to use 6.10.1 for

PR #248, opened 2019-03-21
Fix PMT gain bug in pattern.py

ershockley:
Fixes bug that caused an error for runs where diagnostic PMTs had gains of 0, leading to an `IndexError` when indexing `is_pmt_in`

This needs to be merged ASAP to continue minitree production.


l-althueser:
I think you should use the `self.tpc_channels` [(here)](https://github.com/XENON1T/hax/blob/7e2ceb5509abe487735a7683987db20af68b4e42/hax/treemakers/pattern.py#L50) to get the number of PMTs which belongs to the TPC.

ershockley:
Thanks for the suggestion @l-althueser! Done.

ershockley:
Thanks @JelleAalbers! I'm not sure about the saturated channels - since that's an attribute of the s1 maybe those are by default TPC channel only? 

I'm not authorized to merge with master - could you (or anyone) maybe do it quick?

JelleAalbers:
Ah you're right, saturation is computed on a hit-by-hit basis (https://github.com/XENON1T/pax/blob/master/pax/plugins/peak_processing/BasicProperties.py#L23), but TPC peaks don't have diagnostic hits in them.

PR #249, opened 2019-04-17
update 3d fdc maps for SR2

jingqiangye:
See details in [this note](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenon1t:analysis:sciencerun2:3dfdc).


feigaodm:
will probably update again once we have generated new positions.

PR #258, opened 2019-08-24
Update SR2 TensorFlow Net

weiyuehuan:
This run number updating is for the first 2- month of SR2 data.

The corresponding net can be found in the pax PR:
https://github.com/XENON1T/pax/pull/735


feigaodm:
I'll leave it to @ershockley to check whether the PR is ready.

l-althueser:
Is there a way to know wether this has an impact for certain analyses? Or to phrase it different - should everybody also reload SR0 and SR1 data because it could be different?
This doesn't has to be answered here and now but please keep us updated - Thanks!

feigaodm:
@ershockley So we give up run number 6379?

ershockley:
@l-althueser I think SR0 and SR1 are fine. The runs affected in the SR0-SR1 transition are all earthquake runs so those for sure are fine. The SR1-SR2 transition affects runs that had to be updated anyway. We just need to make sure to always use the gain model to inform these things -- it's also a good double check on the gain model.

ershockley:
@feigaodm it's a bad earthquake run anyway, but according to gain model we'd need another PMT map just for that run right?

feigaodm:
yeah, I understand based on your description, my question is whether will this produce bugs in processing, or anyway another "random" map will be used instead

ershockley:
Ok I see your point. Let's stick it in one of the periods, it will never be used for science data anyway.

feigaodm:
@ershockley Will you make a version_bump in Corrections Minitree version for reprocessing or manually reprocess that period?

ershockley:
@feigaodm if you're okay with it now can you merge?

ershockley:
@feigaodm sorry good timing :) Is the Kr83m electron lifetime model updated as well? If so I think we should do version bump.

PR #261, opened 2019-09-04
Add property for updated s1singlescatter

jingqiangye:
Add several properties as they're necessary for the updated S1SingleScatter, see https://github.com/XENON1T/lax/pull/172


jingqiangye:
Thanks for approving Jelle. Yeah we already used the new properties by generating local minitrees [ExtendedNew](https://github.com/XENON1T/LowER/blob/master/LowER/treemakers.py#L157) (I'm bad with naming..).

PR #265, opened 2020-04-03
Removed all WIMP related blinding cuts

jpienaar13:
Removed all WIMP related blinding cuts in hax as part of unblinding for EFT study (also relevant to Low ER study)


JelleAalbers:
Thanks. I pushed a commit to also remove the double electron capture blinding conditions. These did nothing since they included (... or (run_number>16639) or (run_number<=16639) or ... ).

PR #269, opened 2020-04-20
Added 'S2WithoutAfterpulsePMTs' minitree

DCichon:
This PR aims to add a minitree which contains S2 parameters with contributions from afterpulse PMTs removed. See `xenon:xenon1t:double_beta:cs2_aft_no_ap_pmts` for the cut which motivated it. Code has been tested and compared with the data used in the aforementioned note and yields values which deviate, if at all, at the order of the error induced by the finite float/double precision.


DCichon:
Merging and deploying [this pax PR](https://github.com/XENON1T/pax/pull/739) is required for the minitree to be made.

PR #270, opened 2020-04-26
Missing minitrees and data for 2-fold S1 analysis

feigaodm:
This intend to update the minitrees for AC analysis. This will also open data with r>40cm.

### DIFF
--- a/hax/treemakers/trigger.py
+++ b/hax/treemakers/trigger.py
@@ -189,7 +189,7 @@ class TailCut(hax.minitrees.TreeMaker):
         look_back = 50
         # Load Fundamentals and LargestPeakProperties
         # Using load_single_dataset instead of load will ensure no blindding cut is applies
-        data, _ = hax.minitrees.load_single_dataset(dataset, ['Fundamentals', LoneSignals])#'LargestPeakProperties'])
+        data, _ = hax.minitrees.load_single_dataset(dataset, ['Fundamentals', 'LoneSignals'])#'LargestPeakProperties'])
         if data.empty:
             return pd.DataFrame({})
         # Get largest S2 in the event (or 0, if no S2 was found)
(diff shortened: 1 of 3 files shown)


feigaodm:
The PR has been tested by manual installation. So it shall be ready to go, and doesn't impact existing analysis.

feigaodm:
It would be great if @zhut19 can help check the S2Tails and PreTriggerJunk treemaker
and @ramirezdiego + @jhowl01 can check the LoneSignals and LoneSignalsPreS1. Of course you can check all of them :) Thanks

hax/treemakers/trigger.py
```diff
@@ -181,45 +181,76 @@ def extract_data(self, event):
...
+        t = data['event_time'].values # + data['event_duration'].values/2  #why doesn't it work????
+        tdiff_min = np.zeros(len(t))
+        tdiff_min[0] = 1e18
+        for i in range(1, len(t)):
+            tdiff_min[i] = t[i] - t[i-1]
         # Compute S2/tdif for each value of lookback (from 1 to look_back, inclusive)
         # Allocates n_events * look_back floats; should not be a problem unless look_back is insane number.
         s2_over_tdiff_lookback = np.zeros((len(t), look_back + 1))
         s2_area_lookback = np.zeros((len(t), look_back + 1))
+        se_probabal_coor = np.zeros((len(t), 2, look_back + 1))
```
zhut19, line 217: This will make the first 50 (look back) events have at least a 0 se_probabal_coor, and thus those events would be cut away. Maybe detail in docstring.

feigaodm: It would be good if you can make a fix. But indeed I'm cutting the 1st 100 events in every datasets because we can not calculating their Tail property.

PR #272, opened 2020-08-25
Unblind 0vbb region from 4 to 2 sigma around Qbb

cwittweg:
- Changed blinding from 4 to 2 sigma
- Updated resolution value at Qbb to 0.79 % (see https://arxiv.org/abs/2003.03825)
- Updated g1(z) and g2(z). This is equivalent to making cs1 and cs2b z-dependent as in the published paper (see https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenon1t:analysis:highe_analysis_summary_signal_reconstruction#optimized_energy_resolution)


feigaodm:
Can we assume 0.8% resolution in this analysis, probably we will introduce additional bias, no?It shall have only minor effect, but may affect description in paper.

cwittweg:
> Can we assume 0.8% resolution in this analysis, probably we will introduce additional bias, no?It shall have only minor effect, but may affect description in paper.

It is the published value for the resolution (using the central values of the best fit to the single scatter resolution). In the paper we will take the uncertainty on the energy resolution into account, of course. For the sideband unblinding I can also put 1 % as a more conservative interval. That would make a difference of about 5.2 keV (from 19.4 to 24.6 keV) for one sigma.

feigaodm:
Thanks Christian, I would leave it for you to judge. From my side, 2% might be a better number to justify in the paper. We can always calculate the corresponding sigmas (especially if it's not precisely know yet)

cwittweg:
Thank you. Then let us discuss the final numbers at a later point. Chiara just made me aware that the final value in the EPJC paper is 0.8 %, so for the unblinding I would go with that. Also, we have to change the cs1 and cs2b to the correct cs1_tf and cs2_bottom_ap variables. I missed that these are different, too. We will modify the pull request accordingly.

jpienaar13:
@feigaodm Do you approve the PR?

PR #273, opened 2020-12-28
Unblinding of 0vbb energy region

cwittweg:
Following the decision to unblind the 0vbb data, I removed the respective lines in the code. Please find the link to the unblinding proposal [here](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenon1t:0vbb_unblinding) (@chiaracapelli, @DonNabla, @WolfXeHD).


skazama (approved):

jpienaar13:
@cwittweg Did you test this locally to make sure it behaves as expected on a dataset?

cwittweg:
I didn't check it because the removal is very straightforward. If the line for the blinding is not there, the preselection will not be applied. But if you think it is necessary we can of course check.

jpienaar13:
Just check on a single run if you dont mind, so we dont have to merge things twice

chiaracapelli:
Hi, I just tested it locally. HAX is loaded and here a plot of 100 random datasets during SR1, no cuts applied, only a very loose preselection

unblinding_PR

jpienaar13:
Ok. Seems good

jpienaar13 (approved):

cwittweg:
In any case good that you made us check. I had left in a line that broke this by accident! Lesson learned for the future.

PR #274, opened 2021-01-06
Update pattern.py for blinded regions

chiaracapelli:
Added bypass_blinding to S2PatternReducedAP treemaker

Following unblinding for 0vbb search, the unblinded data resulted blinded again when applying the S2PatternLikelihood cut developed using the S2PatternReducedAP minitrees. A similar issue happened previously in #211. Here the same solution is applied. @cwittweg


chiaracapelli:
Tested on my local HAX installation. No bugs and the issue is solved. In the plot 200 datasets without selections (blue) and after all the 0vbb selections, S2PatternLikelihood included (red).

hax_test

chiaracapelli:
We would need to regrow the S2PatternReducedAP minitrees

jpienaar13 (approved):

